PR #34, opened 2021-04-19
Add a possibility to run/serve the site via Docker

GezimSejdiu:
### Description
This PR provides an easy way to setup/serve the site locally using a light way (Docker) instead of dealing with the dependency hell while installing all the dependencies for using Jekyll.
 
### Issues Resolved
n/a
 
### Check List
- [ ] New functionality includes testing.
  - [ ] All tests pass
- [X] New functionality has been documented.
  - [ ] New functionality has javadoc added

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,19 +14,18 @@ If you discover potential security issues, see the reporting instructions on our
 This project is licensed under the BSD-3-Clause License.
 
 ## Build
+You can build the site and make it available on a local server via: 
+```
+docker-compose up -d
+```
+or by installing all the dependencies on your local environment:
 
 1. Go to the root of the repo
 2. Install [Ruby](https://www.ruby-lang.org/en/) and [RVM](https://rvm.io/)
 3. Install [Jekyll](https://jekyllrb.com/)
 4. Install dependencies: `bundle install`
 5. Build: `bundle exec jekyll serve` for the local server, `bundle exec jekyll build` for one off builds. Either way, the HTML of the site is generated to `/_site`
-6. Point your browser at [`http://127.0.0.1:4000/`](http://127.0.0.1:4000/)
 
-## Docker
-Start a local server via: 
-```
-docker-compose up -d
-```
 and then browse the site at [`http://127.0.0.1:4000/`](http://127.0.0.1:4000/).
 ## Credit
 
--- a/README.md
+++ b/README.md
@@ -20,8 +20,14 @@ This project is licensed under the BSD-3-Clause License.
 3. Install [Jekyll](https://jekyllrb.com/)
 4. Install dependencies: `bundle install`
 5. Build: `bundle exec jekyll serve` for the local server, `bundle exec jekyll build` for one off builds. Either way, the HTML of the site is generated to `/_site`
-6. Point your browser at `http://127.0.0.1:4000/`
+6. Point your browser at [`http://127.0.0.1:4000/`](http://127.0.0.1:4000/)
 
+## Docker
+Start a local server via: 
+```
+docker-compose up -d
+```
+and then browse the site at [`http://127.0.0.1:4000/`](http://127.0.0.1:4000/).
 ## Credit
 
 This website was forked from the BSD-licensed [djangoproject.com](https://github.com/django/djangoproject.com)
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,18 +1,10 @@
-FROM ruby:alpine3.13 as jekyll
-
-RUN apk add --no-cache build-base gcc bash cmake git
-
-RUN gem install bundler -v "~>2.2.11" && gem install jekyll
-
-EXPOSE 4000
-
-ENTRYPOINT [ "jekyll" ]
-
-FROM jekyll/jekyll:4.2.0 as jekyll-serve
+FROM jekyll/jekyll:4.2.0 as jekyll
 
 COPY run.sh /usr/local/bin/
 RUN chmod a+x /usr/local/bin/run.sh
 
 ENTRYPOINT [ "run.sh" ]
+
+EXPOSE 4000
 
 CMD [ "bundle", "exec", "jekyll", "serve", "--force_polling", "-H", "0.0.0.0", "-P", "4000" ]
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -20,8 +20,14 @@ This project is licensed under the BSD-3-Clause License.
 3. Install [Jekyll](https://jekyllrb.com/)
 4. Install dependencies: `bundle install`
 5. Build: `bundle exec jekyll serve` for the local server, `bundle exec jekyll build` for one off builds. Either way, the HTML of the site is generated to `/_site`
-6. Point your browser at `http://127.0.0.1:4000/`
```
dblock, line 23: These instructions seem to imply that I still need to install all the dependencies on my local environment, _then_ run docker. I think it's not a win :) Does it work if I run `docker-compose` as step 2? If so let's make the doc reflect that?

dblock (changes requested):
Cool.

README.md
stockholmux: Agreed.

stockholmux:
(I'm not a Ruby person, excuse my ignorance)

When I run `docker-compose up -d`, the `Gemfile.lock` gets modified and adds `x86_64-linux-musl`. I wouldn't expect _running_ `docker-compose` to modify a file.

In this case, I'm guessing it is just adding a platform dependency so nbd, right? @dblock, you're a Ruby person - care to weigh in?

dblock:
> (I'm not a Ruby person, excuse my ignorance)
> 
> When I run `docker-compose up -d`, the `Gemfile.lock` gets modified and adds `x86_64-linux-musl`. I wouldn't expect _running_ `docker-compose` to modify a file.
> 
> In this case, I'm guessing it is just adding a platform dependency so nbd, right? @dblock, you're a Ruby person - care to weigh in?

When you run `bundle install`, bundler downloads and installs all the dependencies from rubygems and creates Gemfile.lock if it doesn't exist. Commit the file if you want to lock the dependencies.

When you run `docker-compose up -d` it just starts a container. What we want is that the `bundle install` runs *inside that container in its own file system*, which it's currently not doing, so it just is "adding docker on top of an already working `jekyll serve`", which seems like providing a bunch of overhead (docker) to run what otherwise would work today (`jekyll serve`) which doesn't install anything.

Hope I explained it clearly enough? AMA! :)

docker/Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM ruby:alpine3.13 as jekyll
+
+RUN apk add --no-cache build-base gcc bash cmake git
+
+RUN gem install bundler -v "~>2.2.11" && gem install jekyll
```
dblock, line 5: You shouldn't need to `gem install jekyll` (which installs it globally and doesn't use a specific version), it should come in by running `bundle install` (which installs the version declared in Gemfile or locked by Gemfile.lock).

dblock, line 5: Where is Ruby coming from? Parent file system? This container should also fetch and install a recent version of Ruby.

docker/run.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
```
dblock, line 1: Seems redundant, might as well `bundle install` from Dockerfile like other commands.

README.md
GezimSejdiu: Hi @dblock I totally agree with you. I didn't want to remove the steps already preset there so that people may decide if they want to install it natively or via docker.  But I do agree that this is a bit misleading and we should make it clear on the How To on README. Will try to revise it and feel free to comment/suggest.

docker/Dockerfile
GezimSejdiu: Agree and thanks for the catch. Will remove it from it.

GezimSejdiu: I did try to use multistage builds for dividing the ruby (and additional dependencies) and jekyll into two different stages. The Ruby was coming from `ruby:alpine3.13 ` but now it is part of jekyll image: jekyll/jekyll:4.2.0 (latest version of jekyll . Ruby version is bound ([`2.7.0`](https://github.com/envygeeks/jekyll-docker/blob/master/.ruby-version)) to the jekyll docker.

docker/run.sh
GezimSejdiu: You are right. I let this explicitly on the runtime instead of build as an entrypoint. The idea is mostly that we may need to setup a CICD (Github Actions) -- not sure what is the plan here cc:// @stockholmux but we could make use of the same docker image as a PR validation and then instead of building the image always, we can just use any docker registry e.g. Docker Hub to pull the image and then just mount the site to be served.  That workflow will allow (add checks) if changes will break the site or not.

So to recap, we will have a dedicated docker image that will serve as a base for mounting the site, and then this (`/run.sh`) will be executed on the mounted volume instead of rebuilding the image from the beginning.

README.md
GezimSejdiu: I made some changes here https://github.com/opensearch-project/project-website/pull/34/commits/feed774ceb28cf62a563021ae5a622955ae0b48a, but feel free to suggest changes so that it is clear when and how to use `docker-compose`.

stockholmux:
This LGTM - @dblock Any issues before I merge?

dblock (approved):
Much better, thanks. @stockholmux doesn't look like I have permissions to merge so all yours.

docker/run.sh
dblock: FWIW re:Github actions the ruby setup is a 2-liner, and Github actions already run in containers. So using Docker for CI with such a simple setup seems like an unnecessary overhead.

dblock: +1 on enabling some kind of CI to make sure the site builds without markdown errors for example

stockholmux:
Thanks for the contribution, @GezimSejdiu!

PR #42, opened 2021-04-28
Adding JavaDocs interface & added FAQ questions

stockholmux:
Include javadocs interface (no data) and added first batch of FAQ questions


_copyjavadoc.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+
+
+OPENSEARCH_DIR=$1
+CSV_FILE=$2
```
dblock, line 5: Add a comment of what this script does.

Many of these are unused.

dblock (commented):
You might want to move the scripts into a `_script/javadoc` directory or something like that to keep things organized. Add to README how all of this works and what needs to run/when to generate the site. Ideally this should be executed part of jekyll site build, not sure how to do that.

_copyjavadoc.sh
stockholmux: Thanks for point out. should be fixed now.

PR #45, opened 2021-05-04
Organized blog posts into a 2021 folder.

dblock:

### Description

- Consistently named files without `_`.
- Created a 2021 subfolder to plan to avoid a mess in 2022. 
 
Has no impact on what the generated site looks like.


stockholmux:
@dblock the `_` actually is jekyll's way of hiding files. The two blog posts were hidden previously. 

There are several dozen blog posts going back to 2019 that will be pulled over from Open Distro as we deprecate that site. Changing the structure of this would make an already annoying chore even more annoying. I'm glad to re-evaluate the directory structure after porting over all these posts.

PR #46, opened 2021-05-04
Added robots.txt and a sitemap.

dblock:

### Description

Added robots.txt and a sitemap. Let's get that SEO working.


stockholmux:
FYI - I'm holding merging (thus going live) until after beta. I want to reserve the bot crawl for when we have downloadable artifacts :D

dblock:
Currently Google bot indexed things like http://www.opensearch.org/Specifications/OpenSearch/1.1/Draft_5, which looks like we have more problems.

PR #47, opened 2021-05-04
Added a humans.txt with a job search URL.

dblock:

### Description

How do we feel about a humans.txt? Here's [a cool one](https://www.artsy.net/humans.txt), but we don't have any easter eggs, yet. 

### DIFF
--- a/humans.txt
+++ b/humans.txt
@@ -1,0 +1,12 @@
+---
+layout: none
+---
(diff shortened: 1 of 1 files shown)


stockholmux:
Nice! LGTM.

[If someone else wants to throw a URL for an OpenSearch specific job, I think we'd be ok with that too]

PR #50, opened 2021-05-05
Adding OpenSearch FAQs and Beta Announcement Draft to Staging

ahopp:
### Description
Added FAQs (01-12 through 02-03, minus 02/01) and add announcement draft for OpenSearch Beta. 
 
### Issues Resolved
None specifically.


stockholmux:
This looks great! In this case it's fine, but usually it's best to isolate each type change to their own PR, so the beta announcement would in a separate branch. But not a blocker here.

ahopp:
> This looks great! In this case it's fine, but usually it's best to isolate each type change to their own PR, so the beta announcement would in a separate branch. But not a blocker here.

Good callout! I'll isolate better going forward!

PR #56, opened 2021-05-06
Update 01-02-why-was-opensearch-created.markdown

elfisher:
adds missing word "to" between "able" and "use" in second paragraph.


stockholmux:
Thanks! This is a minor change so not a big deal, but we push to 'staging' first usually.

PR #75, opened 2021-05-19
Adds community projects page to website

elfisher:
### Description
This PR adds a community project page similar to the partner page that is self-service so that people can highlight projects they have built for OpenSearch. I have seeded the page with KNN as an example project. Image of the page below.

image


community_projects/index.html
```diff
@@ -0,0 +1,69 @@
...
+<table id="projects" >
+  <tr>
+    <th>Project Name</th>
+    <th>Project Owner</th>
+    <th>Description</th>
+    <th>License</th>
+    <th>Stars</th>
+    <th>Forks</th>
+  </tr>
+  {% for project in site.community_projects %}
```
dblock, line 21: For a static site you probably want to do this as part of the build. It will also work better for SEO.

dblock: https://github.com/18F/jekyll-get does a good job at this

(not sure if we're building the website or relying on GitHub pages to build it in which case custom plugins won't be available)

elfisher: The reason I took this approach is that that builds right now aren't frequent (unless @stockholmux I'm missing something). The GitHub data could get stale (stars, fork counts, etc). Since it is all data coming from GitHub i'm not sure if the tradeoff of putting this into the build process is worth risking having stale data. Thoughts?

elfisher: That makes sense. We don't have many projects listed now, so that's probably not an issue at build time (we could always add batching down the road). I can see if we have many people on the page it could be an issue with the current approach.

dblock: It's more of a problem when bots start hitting the page (e.g. Google bot is very aggressive).

stockholmux: I'm with @dblock - this will be a rate limiting problem. The rate limit will be based on the IP of the user - who knows what they are doing and certainly bots will have an exhausted limit. 

 I think we have a two options:
 1) Use github buttons https://buttons.github.io/
 2) Do this server side with a bash script that creates the yaml files and integrate it into the site build.

dblock: @stockholmux Is this generated with GitHub pages or do you run a jekyll build? If it's the latter, just use that plugin I suggested instead of bash script with yaml.

elfisher:
@dblock and @stockholmux I updated the PR to generate the data at build time with a bash script and updated the webpage accordingly. Please take a look.

community_projects/index.html
elfisher: Unless I'm missing something the plugin only works well if you have **key <> json value** pairs. It doesn't look like it supports **key <> list of json values**. We have the latter; **community_projects <> [project 1, project 2, etc...]**. I took the bash script approach in my latest commit. It was not too complicated, but at build time the `generate.sh` script needs executed to import the data.

dblock:
I like this better than the JS version, except that there's a standard way to invoke something like this in Jekyll during build time, isn't there? I mean there should be a way to remove the need to document to call `generate.sh`.

elfisher:
@dblock I spoke with @stockholmux and he suggested using a script during the build process. I'm not very familiar with the build process for this site so if there is a better way to plug it in I can update, but I might need a pointer to where we run the build steps to do that.

dblock:
Looks like it's actually harder than you'd think :) So just to be clear I think what's in this PR is A-OK.

Now, if you want to raise the bar and do it "the right way", you'd replace the .sh script with a plugin that is invoked every time Jekyll builds the site. There's lots of docs [here](https://jekyllrb.com/docs/plugins/). Leaving it for you as an exercise :)

elfisher:
> Looks like it's actually harder than you'd think :) So just to be clear I think what's in this PR is A-OK.
> 
> Now, if you want to raise the bar and do it "the right way", you'd replace the .sh script with a [generator plugin](https://jekyllrb.com/docs/plugins/generators/) that is invoked every time Jekyll builds the site. There's lots of docs [here](https://jekyllrb.com/docs/plugins/). Leaving it for you as an exercise :)

Sounds good! 😄

stockholmux:
OK. I'm looking at this post RC1. 

@elfisher I like it conceptually, but why are kNN and OpenSearch-cli 'community projects'? They are house in the official org.

elfisher:
> OK. I'm looking at this post RC1.
> 
> @elfisher I like it conceptually, but why are kNN and OpenSearch-cli 'community projects'? They are house in the official org.

@stockholmux my take is they are still community projects even if they are housed in the official org. When we have minimal artifacts this page could serve as a place to identify any community plugin regardless of the github org. I added the github org column to make clear where the project comes from too. I could add all of the other plugins, but was just prioritizing a few to get the page up.

elfisher:
Closing this as it is being addressed in https://github.com/opensearch-project/project-website/pull/153

PR #103, opened 2021-06-01
Upgrading faq

dblock:

### Description

Added upgrading post and FAQ. Merged a couple of questions in. 
Sorted FAQs alphabetically and replaced `#c1` sections with `#names`.
 

dblock:
@elfisher

stockholmux:
@dblock This PR presents a bit of complexity by changing the order and combining questions. The hash link is provided so folks can directly reference parts of the FAQ through link sharing. The changes break this.

For example - Currently:
https://opensearch.org/faq/#q1.9 -> **1.9 Will the OpenSearch API change in future versions?**
After the change:
https://opensearch.org/faq/#q1.9 -> **1.9 Does OpenSearch include forks of Logstash and Beats?**

I believe this also affects the categories, but I'm less worried about this as we don't expose the link to the user as we do for each question.

Can you reformulate the PR to retain the current link structure?

_faqs/01-08-is-opensearch-wire-compatible-with-elasticsearch.markdown
```diff
@@ -1,5 +0,0 @@
----
-category: General
-question: Is OpenSearch wire-compatible with Elasticsearch?
----
-Yes. OpenSearch is a fork of open source Elasticsearch 7.10. As such, it provides the same REST APIs for ingest, search, and management. The query syntax and responses are also the same. In addition, OpenSearch can use indices from Elasticsearch versions 6.0 up to 7.10. We also aim to support the existing Elasticsearch clients that work with Elasticsearch 7.10.
```
elfisher, line 5: Which FAQ is this duplicative of in the ones added? I think we should still highlight that REST APIs remain the same, search APIs remain the same, query responses etc since wire compatibility also has to do with tools being used with OpenSearch.

_faqs/03-04-can-i-upgrade-older-versions-of-elasticsearch-oss-and-kibana-to-opensearch-and-opensearch-dashboards.markdown
```diff
@@ -0,0 +1,7 @@
+---
+question: Can I upgrade older versions of ElasticSearch OSS and Kibana OSS to OpenSearch and OpenSearch Dashboards?
+category: Upgrading
+---
+Elasticsearch OSS and Kibana OSS 5.x up to 6.7.2 can be first upgraded to 6.8.0, then it is recommended to upgrade to Elasticsearch OSS 7.10.2 or ODFE 1.13, before upgrading to OpenSearch and OpenSearch Dashboards.
+
+Note that the minimum supported index version for OpenSearch is 6.0. So, all the 5.x indices have to be re-indexed, before upgrading to OpenSearch.
```
elfisher, line 7: Do we want to bold notes?

_faqs/03-11-can-i-install-elasticsearch-xyz-plugin-in-opensearch.markdown
```diff
@@ -0,0 +1,5 @@
+---
+question: Can I install elasticsearch-xyz-plugin in OpenSearch?
+category: Upgrading
+---
+An ElasticSearch plugin that depends on ElasticSearch JARs will not work without code changes to depend on OpenSearch JARs.
```
elfisher, line 5: ElasticSearch -> Elasticsearch

_faqs/03-12-are-there-any-changes-required-for-my-existing-elasticsearch-clients-to-continue-to-work.markdown
```diff
@@ -0,0 +1,5 @@
+---
+question: Are there any changes required for my existing Elasticsearch OSS clients to continue to work?
+category: Upgrading
+---
+While the OpenSearch API is backwards compatible, some clients or tools may include code, such as version checks, that may cause the client or tool to not work with OpenSearch.
```
elfisher, line 5: If we decide to keep the Wire compatibility FAQ we should update it to include this as a note. It's good to call out.

_posts/2021-06-01-opensearch-backwards-compatibility-faq.markdown
```diff
@@ -0,0 +1,26 @@
...
+
+**Upgrading from Elasticsearch OSS and Kibana OSS or Open Distro for Elasticsearch (ODFE) to OpenSearch and OpenSearch Dashboards is like upgrading between versions of Elasticsearch OSS and Kibana OSS.**
+
+**Specifically:**
+
+* **OpenSearch supports rolling upgrades and restart upgrades from Elasticsearch OSS 6.8.0 through Elasticsearch OSS 7.10.2 to OpenSearch 1.0.**
+* **OpenSearch Dashboards supports restart upgrades from Kibana OSS 6.8.0 through Kibana OSS 7.10.2 to OpenSearch Dashboards 1.0.**
+* **All 1.x versions of ODFE similarly support upgrades to OpenSearch and OpenSearch Dashboards 1.0.**
+
+We've also added an [Upgrading section to OpenSearch FAQs](/faq/#upgrading), and welcome additional questions [on the forums](https://discuss.opendistrocommunity.dev/). Please do open backwards compatibility bugs in [GitHub issues](https://github.com/opensearch-project/OpenSearch/issues).
```
elfisher, line 23: I would including [this issue](https://github.com/opensearch-project/documentation-website/issues/39) in the blog so people can follow our doc updates for upgrading to OpenSearch.

_posts/2021-06-01-opensearch-backwards-compatibility-faq.markdown
```diff
@@ -0,0 +1,26 @@
...
+
+Finally, we would like to note that both spellings of _backward**s** compatibility_ and _backward compatibility_ are allowed, but we prefer _backwards_ because we created GitHub labels as _backwards-compatibility_ and don’t want to go change everything.
```
elfisher, line 25: :)

elfisher (commented):
Added a couple comments and questions. Thx!

dblock:
> @dblock This PR presents a bit of complexity by changing the order and combining questions. The hash link is provided so folks can directly reference parts of the FAQ through link sharing. The changes break this.
> 
> For example - Currently:
> https://opensearch.org/faq/#q1.9 -> **1.9 Will the OpenSearch API change in future versions?**
> After the change:
> https://opensearch.org/faq/#q1.9 -> **1.9 Does OpenSearch include forks of Logstash and Beats?**
> 
> I believe this also affects the categories, but I'm less worried about this as we don't expose the link to the user as we do for each question.
> 
> Can you reformulate the PR to retain the current link structure?

I see. This is actually quite difficult if we are going to move things between categories. The code automatically loads files, then collects them by key, then makes lists that are completely dynamic. I think the only permanent way to do this if we want to support adding more questions is to ensure permalinks based on a field inside the FAQ file. Let me take a stab at that ...

_faqs/03-04-can-i-upgrade-older-versions-of-elasticsearch-oss-and-kibana-to-opensearch-and-opensearch-dashboards.markdown
dblock: I don't think so cause this is not that big of a deal kind of note.

_faqs/01-08-is-opensearch-wire-compatible-with-elasticsearch.markdown
dblock: I've restored this FAQ and provided a more details link.

_faqs/03-11-can-i-install-elasticsearch-xyz-plugin-in-opensearch.markdown
dblock: Roger, will fix elsewhere in the site.

_faqs/03-12-are-there-any-changes-required-for-my-existing-elasticsearch-clients-to-continue-to-work.markdown
dblock: Added.

_posts/2021-06-01-opensearch-backwards-compatibility-faq.markdown
dblock: Done.

dblock:
I've restored the structure and made it more permanent with explicit refs so that we don't depend on sorting. Happy to remove that if you don't think it's necessary @stockholmux, but I think it's clearer this way.

Ready for another pass for a review @elfisher.

elfisher (approved):
LGTM thanks @dblock

dblock:
I don't have permissions to merge here, leaving that to @stockholmux.

_faqs/01-01-what-is-opensearch.markdown
```diff
@@ -1,6 +1,7 @@
 ---
 question: What is OpenSearch?
 category: General
+ref: q1.1
```
stockholmux, line 4: @dblock I don't think using frontmatters to assign the ID is the right take. I had initially considered this method but I went with the auto-assign for two reasons:
1) explicit ID references run the risk of duplicate IDs (breaking linking, also bad practice)
2) this requires some knowledge of what `ref` does - so for folks doing drive-by contributions, it requires a deeper understanding. The filename is more explicit to those who aren't deeply involved. 

I'll admit using filenames as a sorting mechanism seems janky, but I think it's the best option.

stockholmux (changes requested):
Let's go back to filename sorting with IDs assigned by the loop (unless you have a compelling reason otherwise)

_faqs/01-01-what-is-opensearch.markdown
dblock: It's not just janky, it's broken by design without warning. For example, introducing a new category will sort the list of categories differently and renumber everything. You're sure?

dblock:
> Let's go back to filename sorting with IDs assigned by the loop (unless you have a compelling reason otherwise)

I've reverted the explicit numbering and take back what I said on the unpredictable nature of what we have here. I thought that `group_by` was returning categories in random order like typical sets, but it looks like it's predictably returning categories as they appear sorted by filename. So as long as the file names are properly numbered and we don't remove FAQs, the order and numbering should be stable.

stockholmux:
@dblock OK - I've merged this but I'm delaying the blog post included in this PR until we get a bio (see #107). Additionally we have a planned blog post for today (June 3), so we'll slot it un for June 4 assuming we get a bio.

PR #114, opened 2021-06-03
Fix ruby install instructions

dblock:
### Description

Noticed that our instructions for installing ruby here are incorrect. RVM is one option to install Ruby (it manages multiple ruby installations), and if we wanted to recommend RVM we would say "install RVM, then install Ruby". But since YMMV with how you like to install Ruby (e.g via rvm or rbenv), then we should have no opinion.

Added automatic TOC, fixed some minor stuff.
 

stockholmux:
LGTM

PR #137, opened 2021-06-15
docker-compose.yml sample: post OpenSearch-Dashboard to both OpenSearch nodes

ssgao:
### Description
Configure docker-compose.yml to have OpenSearch-Dashboards talk to both OpenSearch nodes.

### Issues Resolved
OpenSearch-Dashboards only talk to one of the two hosts 

### DIFF
--- a/samples/docker-compose.yml
+++ b/samples/docker-compose.yml
@@ -55,13 +55,16 @@ services:
     expose:
       - "5601"
     environment:
-      OPENSEARCH_HOSTS: https://opensearch-node1:9200
+      OPENSEARCH_HOSTS: '["https://opensearch-node1:9200","https://opensearch-node2:9200"]'
     networks:
       - opensearch-net
 
 volumes:
   opensearch-data1:
+    driver: local
   opensearch-data2:
+    driver: local
 
 networks:
   opensearch-net:
+    driver: bridge
--- a/samples/docker-compose.yml
+++ b/samples/docker-compose.yml
@@ -61,10 +61,7 @@ services:
 
 volumes:
   opensearch-data1:
-    driver: local
   opensearch-data2:
-    driver: local
 
 networks:
   opensearch-net:
-    driver: bridge


elfisher:
@ssgao I'm having trouble reproducing this issue on my machine (I'm using an x86 mac). The current docker image on the website works for me without these updates. Could you share more details about what might be causing the issue?

ssgao:
@elfisher 
It is strange. It was causing problems but now when I remove them I don't have a problem starting up.

Do you think the OpenSearch hostname change is needed? If not, I'll close the PR.

stockholmux:
I don't believe this is required, although I'm not an expert on Docker Compose. This was sourced from the documentation - @aetter do you have thoughts?

ssgao:
@stockholmux definitely not required.
I added it here because when I was playing the docker file with a 4 node setup and taking node up and down to test in-sync allocations and consensus. I expected OpenSearch-Dashboards to be able to talk to all OpenSearch nodes. Here it would only talk to node1, and if node1 is down, it won't work anymore.
I struggled a bit to figure out the correct syntax to put here so OpenSearch-Dashboards (really hope we get a abbreviation for this btw, it's a lot of keystrokes lol) node would talk to other nodes in case node1 goes down.

elfisher:
> @elfisher
> It is strange. It was causing problems but now when I remove them I don't have a problem starting up.
> 
> Do you think the OpenSearch hostname change is needed? If not, I'll close the PR.

@ssgao my instinct is to keep this file simple since it is intended for getting started. If it now works on your side. I think we should close this PR. However, I did notice your comment about node1 being down and causing issues. Did you just kill the node1 container and then see what happened? I can give that a try.

ssgao:
@elfisher 
Exactly.
In all fairness, with a 2 node cluster, killing a node would violate minimum consensus requirement, so OpenSearch won't progress regardless, but I think the docker-compose file could do a better job by demonstrating how can OpenSearch-Dashboards talk to multiple nodes.

elfisher:
> @elfisher
> Exactly.
> In all fairness, with a 2 node cluster, killing a node would violate minimum consensus requirement, so OpenSearch won't progress regardless, but I think the docker-compose file could do a better job by demonstrating how can OpenSearch-Dashboards talk to multiple nodes, so later if user choose to expand this setting into a 3 node cluster, they won't have to struggle with this problem.

That does make sense. Just adding the second host fixes that?

elfisher:
@aetter if we update this, we probably want to do something similar in the one in our documentation website. I'll cut a PR for that when we come to a conclusion here.

stockholmux:
I'm guessing its 2 nodes because 2 nodes is the minimal 'cluster' one can play. The resources to run 2 vs 4 nodes is quite a bit larger, right? It seems like the resource usage is already quite high with 2 nodes for someone trying it out over docker on a laptop.

I wouldn't be opposed to setting up multiple docker compose files though

ssgao:
@stockholmux
I think we are talking about two different issues here.
Your point of having multiple docker-compose files to represent different setups is completely valid.

My point here is simply to have OpenSearch-Dashboards talk to both nodes, since there are two nodes there available to serve traffic anyways.

elfisher:
> @stockholmux
> I think we are talking about two different issues here.
> Your point of having multiple docker-compose files to represent different setups is completely valid.
> 
> My point here is simply to have OpenSearch-Dashboards talk to both nodes, since there are two nodes there available to serve traffic anyways.

I tested the setup and took down node 1. This compose works as outlined by @ssgao. I think we should merge this. I also opened a PR in the docs https://github.com/opensearch-project/documentation-website/pull/54

stockholmux:
@ssgao / @elfisher - my point in suggesting multiple docker-compose files is a reaction to this:

_In all fairness, with a 2 node cluster, killing a node would violate minimum consensus requirement, so OpenSearch won't progress regardless_

What does having two nodes in a cluster gain us if we're just specifying those nodes explicitly?

ssgao:
@stockholmux 
I understand your point, and I do think that having another 3 or maybe 4 node setup docker-compose file can help the user  play with various down scenarios.
IMO, 4-node is actually the most ideal way to learn about OpenSearch behavior locally since with 3 of them being master-eligible, you need 2 nodes minimum to inspect `_cluster/state`, so it's really just bringing up and down 2 other nodes and see how _in-sync allocations_ works and how global & local checkpoint works, etc.
Having that said, a 2-node docker-compose file is still useful so the user can still start the cluster without changing default 2GB docker desktop memory setting in [Mac](https://docs.docker.com/docker-for-mac/#resources) and [Windows](https://docs.docker.com/docker-for-windows/#resources). Trying to start 3-node cluster without changing this setting would result in OOM error and causing nodes to crash.
Alternatively, to keep 3-node cluster, we can lower the heap size from 512MB to 256MB. I tried it and it actually works. The stack still boots. I have not done any load testing on it so I'm not sure about the usability of 3 256MB OpenSearch node cluster, but that is still a possibility.

Now, I'm not 100% clear about your question
> What does having two nodes in a cluster gain us if we're just specifying those nodes explicitly?

Are you asking why we should specify two OpenSearch nodes in the `OPENSEARCH_HOSTS` setting?
I think I have explained it a few times, but bear with me again, let me try with an analogy.
Say we have a load balancer, and there's one node behind the load balancer serving the traffic
Now we add another node to serve traffic, but we don't register to the load balancer.
Load balancer still directs all traffic to the old node.
Do you see an issue with this setting?

stockholmux:
@ssgao No, I get that. I was thinking from the perspective of someone creating a _n_ node cluster with compose. It would be a bit weird to explicitly state each node `OPENSEARCH_HOSTS` right? 

My point is that by specifying a 2 node cluster with explicitly defined nodes are we giving people a docker-compose file that is an anti-pattern or unusual configuration.

ssgao:
@stockholmux 
Ah I see.
Honestly, I think this could be an enhancement request to allow OpenSearch-Dashboards to discover all coordinator-eligible OpenSearch nodes.
However, to my best knowledge, users today just hard-code these node urls in kibana.yml file.

elfisher (approved):
This LGTM. From what I can tell there isn't a way to auto discover all nodes via a OpenSearch Dashboards yml config. I suggest we move forward with this update as it demonstrates how to route Dashboards to multiple nodes without introducing significant complexity.

stockholmux (approved):
OK. Let's do this!

dblock:
I wonder whether we need a maintained repo for docker samples, ala https://github.com/deviantony/docker-elk/

PR #139, opened 2021-06-15
updates the upgrade FAQ section title to be more specific 

elfisher:
updates the upgrade FAQ section title to specifically say "Upgrading to OpenSearch." _Upgrading_ might be perceived as upgrading from OpenSearch 1.x to 1.y. However, these FAQs are focused upgrading from open source ES and ODFE to OpenSearch.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -71,7 +71,7 @@ GEM
 PLATFORMS
   ruby
   x86_64-darwin-19
-  x86_64-linux-musl
+  x86_64-linux
 
 DEPENDENCIES
   jekyll (~> 4.2.0)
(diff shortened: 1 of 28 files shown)


stockholmux:
Will we create a separate section for upgrading between versions in the future?

elfisher:
> Will we create a separate section for upgrading between versions in the future?

I think we would. There is a lot of content specific to upgrading to OpenSearch and I suspect the FAQs between OpenSearch versions might be different (for example the differences between Major, Minor, and Patch updates would only be relevant to OpenSearch version upgrades). I wouldn't want to overload the upgrade from open source ES and ODFE to OpenSearch topic.

stockholmux (approved):
LGTM

PR #144, opened 2021-06-22
Visual updates to partners page

kgcreative:
### Description
These are mainly cosmetic CSS and markup changes to the partners page. I'm keeping the information and hierarchy the same, just adding some visual polish + responsive styles.
 
### Before
Screen Shot 2021-06-22 at 3 13 19 PM

### After
Screen Shot 2021-06-22 at 3 12 30 PM

### DIFF
--- a/_layouts/fullwidth.html
+++ b/_layouts/fullwidth.html
@@ -4,6 +4,6 @@ layout: default
 
 {% assign primary_title = page.primary_title %}
 {% include copy_banner.html %}
-<div class="container">
+<main class="container">
     {{ content }}
-</div>
+</main>
--- a/_sass/_opensearch.scss
+++ b/_sass/_opensearch.scss
@@ -168,19 +168,6 @@ dl.list-features {
     }
 }
 
-#partners {
-    .partner {
-        text-align: center;
-        float: left;
-
-        width: 100px;
-        img {
-            height: 60px;
-            width: 60px;
-        }
-    }
-}
-
 .img-fluid {
     max-width: 100%;
     height: auto;
--- a/_sass/_partners.scss
+++ b/_sass/_partners.scss
@@ -14,14 +14,12 @@
 
     @include respond-min(580px) {
         display: grid;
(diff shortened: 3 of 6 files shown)


elfisher:
This LGTM. @kgcreative are you thinking we can expand on this layout so that partners could add quotes or other content in each card?

kgcreative:
@elfisher Yes! I need to reason about what's the longest and shortest possible content that could go in a card, and then we might need to switch the layout to 2 or 3 columns instead of 4 max, but either way, this format should be able to support a richer information display

elfisher (approved):
LGTM

stockholmux:
@kgcreative Code looks good. Nice and clean. 

Visually, I'm struggling a bit:
- I can't say I'm really wild about the drop shadow (we don't use this style anywhere else on the site).  Could we darken or widen he border instead for a hover effect?

- The hover animation on the text is something I can't quite decide if I like. It's really smooth and nice, but it's bit noticeable. 

Also, when these have more content in them I think we'll need to make the click area button rather than the whole thing.

kgcreative:
> * I can't say I'm really wild about the drop shadow (we don't use this style anywhere else on the site).  Could we darken or widen he border instead for a hover effect?

I can do that. I wanted to float the ideal of the more subtle style with drop shadows in more places on the site, but perhaps I will leave that for a later PR that addresses styles a bit more globally

> * The hover animation on the text is something I can't quite decide if I like. It's really smooth and nice, but it's bit noticeable.
Similar as above, i would like to promote this for most links. Perhaps this is a change I can introduce at a more global level as part of a CSS style override later on.

> Also, when these have more content in them I think we'll need to make the click area button rather than the whole thing.

Yes. The intent here is for these to have the wider click target until we have more content, at which point we can change the link to a button, and make the just button be the clickable element.

elfisher:
@kgcreative does this open up the partner page to support non-square logos? I know some businesses have rectangular logos, so this might make it easier to include those.

kgcreative:
@elfisher It does. There's no reason we should limit to square. I'll test some edge cases (long and skinny, tall and short) to make sure we have good boundaries around max sizes. It'll be with the ammended commit once I address Kyle's concerns above.

kgcreative:
@stockholmux - Updated branch per your comments above.

stockholmux (approved):
LGTM

PR #214, opened 2021-07-21
removed unneeded JS dependencies, updated JS dependencies

stockholmux:
### Description
A number of JS dependencies were woefully out-of-date. This fixes a number of issues picked up by an automated dependency scanner.

1) Updated jquery to the latest (3.6) from 1.11.2 
2) Removed `jquery.inview`, `jquery.flot`, and `jquery.unveil` as they appear to be unused (remnant from forked project)
3) Removed demo code from webfontloader as it was packaging a even older version of jquery that was being picked up by a dependency scanner.
4) Removed features being checked for but not used in this project


assets/js/lib/jquery/dist/jquery.js
```diff
@@ -1,10346 +0,0 @@
-/*!
- * jQuery JavaScript Library v1.11.2
```
tmarkley, line 2: I've never used Jekyll, but is there not a configuration file that specifies the dependencies (and their respective versions) for the application?

assets/js/main.js
```diff
@@ -34,9 +34,9 @@ define(function() {
     }
 
     //feature list
-    if (hasClass('list-features')) {
+    /*if (hasClass('list-features')) {
         mods.push('mod/list-feature');
-    }
+    }*/
```
tmarkley, line 39: `Removed features being checked for but not used in this project`

Can we just delete these lines then?

assets/js/lib/jquery/dist/jquery.js
stockholmux: Jekyll is purely server-side - it isn't opinionated in front-end dependencies. 

This project _did_ use bower for front-end package management (inherited from the forked project). Bower is functionally deprecated and not even very useful in the simplified scope of the current project. After this is resolved (for security), the next challenge is moving to a more future-proof package management.

assets/js/main.js
stockholmux: Makes sense.

assets/js/lib/jquery/dist/jquery.js
tmarkley: Okay that makes more sense - I saw the bower config but didn't see it updated here so I wasn't sure. Do we have an issue opened to address package management?

stockholmux: Good idea. I added #221

tmarkley (commented):
Last question - what's the testing strategy? How do we confirm that nothing is broken? I didn't see anything listed in the repo.

stockholmux:
@tmarkley Test is fair call out. Now and before this change, it's been manual testing. Since 99% of the project is static HTML, if a quick review in-browser works, we're pretty safe moving forward. Its _mostly_ just content in templates.

That being said, it's probably a good idea to get more serious about testing, especially as the complexity will continue to grow over time

tmarkley (approved):
LGTM!

PR #256, opened 2021-08-11
Sitemap enhancements & URL fixes

AMoo-Miki:
Signed-off-by: Miki <mehranb@amazon.com>

### Description
* Improves sitemap's last-modified
* Fixes paths to `versions`, `artifacts`, and `authors`
* Lets robots know to not bother with `javadocs`
 
### Issues Resolved

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -17,6 +17,7 @@ gem "minima", "~> 2.5"
 group :jekyll_plugins do
   gem "jekyll-feed", "~> 0.12"
   gem "jekyll-paginate"
+  gem "jekyll-last-modified-at"
   gem "jekyll-sitemap"
 end
 
--- a/_config.yml
+++ b/_config.yml
@@ -16,6 +16,7 @@ headings:
 plugins:
   - jekyll-feed
   - jekyll-paginate
+  - jekyll-last-modified-at
   - jekyll-sitemap
 exclude:
   - Gemfile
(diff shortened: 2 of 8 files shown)


_config.yml
```diff
@@ -54,13 +55,13 @@ defaults:
       type: versions
     values: 
       layout: versions
-      permalink:  versions/:categories/:slug
+      permalink:  versions/:categories/:slug.html
```
stockholmux, line 58: Good catch on these @AMoo-Miki.  Is there an SEO advantage to using raw paths (`/versions/foo/bar/`) vs files (`/versions/foo/bar.html`)? The site is pretty inconsistent when I step back and look at it.

Gemfile
```diff
@@ -17,6 +17,7 @@ gem "minima", "~> 2.5"
 group :jekyll_plugins do
   gem "jekyll-feed", "~> 0.12"
   gem "jekyll-paginate"
+  gem "jekyll-last-modified-at"
```
stockholmux, line 20: what does this plugin do? Why is it needed?

stockholmux (commented):
A few questions...

_config.yml
AMoo-Miki: SEO-wise, none. The only reason I chose the `.html` was because the permalinks that visitors might have shared.

stockholmux: Makes sense - given the name of the PR branch, I just wanted to confirm that we weren't doing anything counter to SEO best practices.

Gemfile
stockholmux: OK. Makes sense - if it helps the site map 👍

stockholmux (approved):
LGTM

PR #321, opened 2021-09-20
Add FreeBSD installation instructions

smortex:
FreeBSD ports for OpenSearch and OpenSearch-Dashboards have been
committed to the FreeBSD ports tree, and packages will be available soon
to FreeBSD users through the regular package management system.

Add FreeBSD to the list of platforms.  Because installation instruction
do not change based on the architecture, add a "All" architecture to
avoid duplicating the same information multiple times.

Also include a link to the port page on FreshPorts which allow to
"conveniently" see the port history (in a more convenient way than
cgit).

While here, adjust the list of all artifacts to order them
alphabetically in the artifacts directory.

### DIFF
--- a/_artifacts/opensearch-1.0/all/opensearch-1.0.1-freebsd-all.markdown
+++ b/_artifacts/opensearch-1.0/all/opensearch-1.0.1-freebsd-all.markdown
@@ -1,0 +1,12 @@
+---
+role: daemon
+artifact_id: opensearch
+version: 1.0.1
+platform: freebsd
+architecture: all
+slug: opensearch-1.0.1-freebsd-all
+category: opensearch
+type: pkg
+freebsd_package_name: opensearch
+link: https://www.freshports.org/textproc/opensearch
+---
--- a/_artifacts/opensearch-1.0/all/opensearch-1.0.1-freebsd-all.markdown
+++ b/_artifacts/opensearch-1.0/all/opensearch-1.0.1-freebsd-all.markdown
@@ -6,7 +6,7 @@ platform: freebsd
 architecture: all
 slug: opensearch-1.0.1-freebsd-all
 category: opensearch
-type: pkg
+type: system-package
 freebsd_package_name: opensearch
 link: https://www.freshports.org/textproc/opensearch
 ---
--- a/_artifacts/opensearch-1.0/all/opensearch-dashboards-1.0.1-freebsd-all.markdown
+++ b/_artifacts/opensearch-1.0/all/opensearch-dashboards-1.0.1-freebsd-all.markdown
@@ -1,0 +1,13 @@
+---
+role: ui
+title: OpenSearch Dashboards
+artifact_id: opensearch-dashboards
+architecture: all
+platform: freebsd
+type: pkg
+category: opensearch-dashboards
+slug: opensearch-dashboards-1.0.1-freebsd-all
+version: 1.0.1
+freebsd_package_name: opensearch-dashboards
+link: https://www.freshports.org/textproc/opensearch-dashboards
+---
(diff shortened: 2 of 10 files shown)


smortex:
Screenshot from the rendering of the PR:

image

_artifacts/opensearch-1.0/all/opensearch-1.0.1-freebsd-all.markdown
```diff
@@ -0,0 +1,12 @@
+---
+role: daemon
+artifact_id: opensearch
+version: 1.0.1
+platform: freebsd
+architecture: all
+slug: opensearch-1.0.1-freebsd-all
+category: opensearch
+type: pkg
```
stockholmux, line 9: Humm. `pkg` is a bit overloaded as it's used in macos too.

_artifacts/opensearch-1.0/all/opensearch-dashboards-1.0.1-freebsd-all.markdown
```diff
@@ -0,0 +1,13 @@
+---
+role: ui
+title: OpenSearch Dashboards
+artifact_id: opensearch-dashboards
+architecture: all
+platform: freebsd
+type: pkg
```
stockholmux, line 7: see above. I think it might need some disambiguation.

_layouts/versions.html
```diff
@@ -46,11 +48,13 @@
   - x64
   - arm64
   - x86
+  - all
```
stockholmux, line 51: would `noarch` work better than all? I've not settled on this... 🤔

_artifacts/opensearch-1.0/all/opensearch-1.0.1-freebsd-all.markdown
smortex: Quite agree… Maybe something like 'system-package' to insist on the way this is managed using the regular package manager of the system?

_layouts/versions.html
smortex: `noarch` make me think "platform agnostic" (e.g. we use `noarch` to refer to FreeBSD packages that are identical on all architectures, e.g. python / ruby).  Here we indeed have platform-dependent packages because the application bundle architecture specific binaries (jna for OpenSearch; re2 for OpenSearch-Dashbaords).

Maybe `any` is a better fit?

smortex:
As of today, `textproc/opensearch` and `textproc/opensearch-dashboards` packages are available for FreeBSD 11, 12, 13 and 14 for both amd64 and i386 (except FreeBSD 12 amd64 where a build dependency seems to be broken).

The packages availability matrix is visible in FreshPorts at the URI the PR associate with the artifacts.

dblock:
🚀

_layouts/versions.html
stockholmux: @smortex OK - so what architectures are supported in the FreeBSD version?

smortex: @stockholmux we currently have i386, amd64 (x86, 32 and 64 bits) and aarch64 (arm 64 bits) packages available.  Other platforms might not have packages because of a failing dependency (I am not sure openjdk11 build on 32 bit ARM) or be delayed  because of limited build resources (I only saw aarch64 packages today).

stockholmux: OK. That makes sense @smortex I think it makes sense to more explicit here (repeating near identical `/_artifact/` files) than less (using `all` or `any`). Two reasons:
1) I'd like these files to be useful to eventually to power an API and making sure there is an explicit entry for each architecture would simplify the the implementation.
2) Who knows, maybe someone ports OpenSearch to RISC-V (or whatever) and would be frustrated `all` not being true.

smortex: Sure!  How should we name 32bit x86 (`i386` architecture with FreeBSD terminology)? `x86` (what Microsoft [seems to use for this architecture](https://docs.microsoft.com/en-us/windows-hardware/drivers/debugger/x86-architecture))?

smortex: Replying to myself: `x86` is not listed in the _architectures_ list but appear in the _architecture_order_ so should be fine :-)

stockholmux:
Scheduled for production on 2021/10/4

PR #434, opened 2021-11-01
adds more testimonials

elfisher:
### Description
Adds additional testimonials to the testimonial page.


elfisher:
@stockholmux some of the logos aren't rendering to similar sizes. let me know if you have suggestions on how we can fix that.

stockholmux (changes requested):
It looks good, however the images need a little work.
- wipro is _huge_ at 188kb and 2560x2560. That's 10x file size of Goldman and double the resolution. It should be resized - really no need for the file to be that high res. On screen it's taking up 54 px square. We should over size these images to account for retina displays, but not 47x!
- wipro has a ton of padding built into the image which makes it look small when scaled - when I overlay goldman and wipro, you can see the difference. overlay. 
If you crop the wipro logo closer, they will appear closer in size

elfisher:
> It looks good, however the images need a little work.
> 
> * wipro is _huge_ at 188kb and 2560x2560. That's 10x file size of Goldman and double the resolution. It should be resized - really no need for the file to be that high res. On screen it's taking up 54 px square. We should over size these images to account for retina displays, but not 47x!
> * wipro has a ton of padding built into the image which makes it look small when scaled - when I overlay goldman and wipro, you can see the difference. 
>   If you crop the wipro logo closer, they will appear closer in size

Fixed accordingly. Looks better!

stockholmux (approved):
LGTM

PR #540, opened 2021-11-30
Add custom search indexing and UI

AMoo-Miki:
### Description
* Add a plugin that produces indexible content: This content is fed to OpenSearch and made available via https://search-api.opensearch.org/search
* Add a search text input, identical to the one used on documentation
* Converted some `768px` media queries to `46.25rem` which is `740px` (matching [defaults from just-the-docs](https://pmarsceill.github.io/just-the-docs/docs/utilities/responsive-modifiers/) so the website and docsite appear uniform)
* Set `versions` landing page to `omit_from_search: true`
* Moved the title of `source` from config to the page itself as it was used once and helps simplify indexing
 

### Check List

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -7,7 +7,6 @@ twitter_username:
 github_username: 
 headings:
   news: 'News'
-  source: 'Source Code'
   authors: 'Authors'
   downloads: 'Download & Get Started'
 
--- a/_plugins/search-indexer.rb
+++ b/_plugins/search-indexer.rb
@@ -1,0 +1,124 @@
+# frozen_string_literal: true
+
+require "jekyll/hooks"
+require "jekyll/document"
+require "json"
+
+##
+# This singleton facilitates production of an indexable JSON representation of the content to populate a data source
+# to provide search functionality.
+
+module Jekyll::ContentIndexer
+
+  ##
+  # The collection that will get stores as the output
+
+  @data = []
+
+  ##
+  # Pattern to identify documents that should be excluded based on their URL
+
+  @excluded_paths = /(^\/blog\/(page\d|$)|^\/events\/([^\d]|$)|^\/faqs|\.(css|js|json|map|xml|txt|yml)$)/i.freeze
+
+  ##
+  # Pattern to identify block HTML tags (not comprehensive)
+
+  @html_block_tags = /\s*<[?\/]?(article|blockquote|d[dlt]|div|fieldset|form|h|li|main|nav|[ou]l|p|section|table|t[rd]).*?>\s*/im.freeze
+
+  ##
+  # Pattern to identify certain HTML tags whose content should be excluded from indexing
+
+  @html_excluded_tags = /\s*<(head|style|script|h1).*?>.*?<\/\1>/im.freeze
+
+  ##
+  # Initializes the singleton by recording the site
+
+  def self.init(site)
+    @site = site
+  end
+
+  ##
+  # Processes a Document or Page and adds it to the collection
+
+  def self.add(page)
+    return if @excluded_paths.match(page.url)
+
+    content = page.content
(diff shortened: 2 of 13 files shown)


_config.yml
```diff
@@ -7,7 +7,6 @@ twitter_username:
 github_username: 
 headings:
   news: 'News'
-  source: 'Source Code'
```
stockholmux, line 10: why tho?

versions/index.html
```diff
@@ -1,5 +1,6 @@
 ---
 layout: fullwidth
+omit_from_search: true
```
stockholmux, line 3: is this the way that you prevent a page being indexed? We should document this somewhere

_plugins/search-indexer.rb
```diff
@@ -0,0 +1,136 @@
...
+  # Initializes the singleton by recording the site
+  def self.init(site)
+    @site = site
+  end
+
+  ##
+  # Processes a Document or Page and adds it to the collection
+  def self.add(page)
+    return if @excluded_paths.match(page.url)
+    return if page.data['omit_from_search']
```
stockholmux, line 42: does `page` also work for layouts? So, say, I want any page with a given layout to _not_ show up in search. Would this work?

stockholmux (commented):
A few confusing points.

_config.yml
AMoo-Miki: Initially, I needed a less complicated way to map titles for pages that didn't have one without having to look at `headings` and try to map it based on the file name, or part of url. I noticed that pages like `/authors` worked fine because they had `primary_title` in their front matter; I added `primary_title` to the `source.markdown` file. However, since that was only place that `headings.source` was used, there was no longer a need to keep it.

versions/index.html
AMoo-Miki: Yes it is. I added it to the indexer plugin's inline documentation.

stockholmux: I think we should be more explicit than that - can we throw it in the `README.md` file?

_config.yml
stockholmux: OK. I think the original idea was to have only one place to have to internationalize, but if that's not how we ended up involving in practice, I guess I can roll with it.

_plugins/search-indexer.rb
AMoo-Miki: `page.data` is populated the particular document's front matter after its content has been converted but before before being rendered in a layout. Hence, setting `omit_from_search: true` will not work. However, `@excluded_paths` is a hard-coded regex and can be used for this purpose.

stockholmux: OK. In the templating language for Jekyll, `page` only refers to front matter directly on the page but there are similar objects for layouts (`layout`). I'll experiment at a later - specifying it up the stream might be very useful.

stockholmux (dismissed):
LGTM

_plugins/search-indexer.rb
AMoo-Miki: The hook being used is `post_convert` to avoid having to remove content added by layouts. I might be able to look check `omit_from_search` after rendering and remove documents from the index. I will experiment with that for the next iteration.

_plugins/search-indexer.rb
```diff
@@ -0,0 +1,139 @@
+# frozen_string_literal: true
```
dblock, line 1: Where's this code from? Written from scratch? Add a license header?

stockholmux: good catch @dblock. @AMoo-Miki I think this is original so it needs a headering.

AMoo-Miki: It is from scratch; will add header.

stockholmux (approved):
LGTM

PR #610, opened 2022-01-04
Add the ContentModifier plugin to mitigate security risks automatically and prioritize plugins

AMoo-Miki:
Signed-off-by: Miki <miki@amazon.com>

### Description
* Add the ContentModifier plugin to mitigate security risks automatically
* Add reverse tabnabbing mitigation to ContentModifier
* Add execution priority to the existing plugins
* Prevent unnecessary execution of hooks when the Serve command is used and provide a way to override this behavior
* Optimize hook registration

### Check List

### DIFF
--- a/_plugins/content-modifier.rb
+++ b/_plugins/content-modifier.rb
@@ -1,0 +1,102 @@
+# frozen_string_literal: true
+
+# Copyright OpenSearch Contributors
+# SPDX-License-Identifier: BSD-3-Clause
+
+require "jekyll/hooks"
+
+##
+# This singleton modifies the content at build time. Its primary use is to mitigate security risks automatically.
+#
+# The plugin does not apply changes when the Serve command is used but `JEKYLL_ALLOW_CONTENT_MODIFIER`, set on the
+# environment, will override the behavior.
+# Usage: `JEKYLL_ALLOW_CONTENT_MODIFIER= bundle exec jekyll serve --trace`
+
+module Jekyll::ContentModifier
+
+  ##
+  # Pattern to identify documents that should be excluded based on their URL
+  @excluded_paths = /\.(css|js|json|map|xml|txt|yml|svg|)$/i.freeze
+
+  ##
+  # Defines the priority of the plugin
+  # The hooks are registered with a very high priority to make sure the changes are in before other hooks are run
+  def self.priority
+    90
+  end
+
+  ##
+  # Initializes the singleton by recording the site
+  def self.init(site)
+    @site = site
+
+    # Avoid initializing if serving and not forced to run
+    if site.config["serving"] and (!ENV.key?('JEKYLL_ALLOW_CONTENT_MODIFIER') or ENV['JEKYLL_ALLOW_CONTENT_MODIFIER'] == "false")
+      return Jekyll.logger.info "ContentModifier:",
+                                "disabled. Enable with JEKYLL_ALLOW_CONTENT_MODIFIER on the environment"
+    end
+
+    # Process a Page as soon as its content is ready
+    Jekyll::Hooks.register :pages, :post_convert, priority:self.priority do |page|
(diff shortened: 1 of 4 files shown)


stockholmux:
Interesting - I have to ask if this is needed since [tab nabbing has been mitigated at the browser level for quite some time](https://www.zdnet.com/article/chrome-to-block-tab-nabbing-attacks/).

AMoo-Miki:
> Interesting - I have to ask if this is needed since [tab nabbing has been mitigated at the browser level for quite some time](https://www.zdnet.com/article/chrome-to-block-tab-nabbing-attacks/).

I agree that it has been blocked by major browsers for a year now but I also think a year is too short a time to forget about it. I also worry about derivative browsers that might have not pulled in the changes.

stockholmux:
OK - A couple of things:
- This plugin increases a cold build time from avg 9.028 second to 11.231 (+~24%) and subsequent builds by 7.988 seconds to  8.824 (+~10.4%). Build times are critical DX for working on the site.
- This method will continually impact build times as the site grows.
- [IE11 is EoL on June 22, 2022](https://docs.microsoft.com/en-us/lifecycle/faq/internet-explorer-microsoft-edge) (Hooray!), so if we do add this plugin, it should only really be needed until that date.

Another option would be do this through JS only on browsers that don't properly mitigate tabnabbing.

AMoo-Miki:
> OK - A couple of things:
> 
>     * This plugin increases a cold build time from avg 9.028 second to 11.231 (+~24%) and subsequent builds by 7.988 seconds to  8.824 (+~10.4%). Build times are critical DX for working on the site.
> 
>     * This method will continually impact build times as the site grows.
> 
>     * [IE11 is EoL on June 22, 2022](https://docs.microsoft.com/en-us/lifecycle/faq/internet-explorer-microsoft-edge) (Hooray!), so if we do add this plugin, it should only really be needed until that date.
> 
> 
> Another option would be do this through JS only on browsers that don't properly mitigate tabnabbing.

I added mechanism to not run this and the ContentIndexer plugins while developing using `serve`. This should eliminate the burden on contributors.

I did a quick check for the website visitors during the last month and only 1.38% to 1.67% of them were using vulnerable browsers. While those numbers are way too low already, if we use the plugin only for a few months, Android Browser will catch up and IE11 will be EoLed, slashing the vulnerable users down to 0.6%.

_plugins/content-modifier.rb
```diff
@@ -0,0 +1,98 @@
+# frozen_string_literal: true
```
stockholmux, line 1: Does this need a `SPDX-License-Identifier` and a copyright?

_plugins/content-modifier.rb
```diff
@@ -0,0 +1,98 @@
...
+    90
+  end
+
+  ##
+  # Initializes the singleton by recording the site
+  def self.init(site)
+    @site = site
+
+    # Avoid initializing if serving and not forced to run
+    if site.config["serving"] and (!ENV.key?('JEKYLL_ALLOW_CONTENT_MODIFIER') or ENV['JEKYLL_ALLOW_CONTENT_MODIFIER'] == "false")
```
stockholmux, line 30: Can we drop a line or two into README about how to turn it on and off (and what it does)?

stockholmux (changes requested):
A couple of nits and I'm fine with this!

stockholmux:
@AMoo-Miki Any movement on the changes requested?

_plugins/content-modifier.rb
AMoo-Miki: Done.

_plugins/content-modifier.rb
AMoo-Miki: Added.

stockholmux (approved):
LGTM

PR #626, opened 2022-01-17
V1 2 4

stockholmux:
### Description
- Adds version file for 1.2.4
- Adds artifacts files for 1.2.4  (freebsd artifacts are hidden with `_`)
- Adds automation to generate versions files
- Renames artifact automation script to hide from Jekyll

### Issues Resolved
n/a

### Check List


peterzhuamazon:
@stockholmux @elfisher Are we supposed to release freebsd version only after other artifacts are out to public?
Should we remove the freebsd files until then?
Thanks.

stockholmux:
@peterzhuamazon The FreeBSD folks need to take some actions before a version is available. In this update, the FreeBSD files are added, but they are hidden from rendering (`_` at the start of a file is ignored by Jekyll).

So, when FreeBSD is ready, they just need to rename the file.

peterzhuamazon (changes requested):
Thanks.

peterzhuamazon (approved):
Thanks.

PR #642, opened 2022-01-22
Add Arch Linux artifacts

hashworks:
### Description
Adds artifacts for the official [Arch Linux OpenSearch packages](https://archlinux.org/packages/?q=opensearch).

As discussed with @stockholmux by mail.

The Arch Linux packages I created do stuff a bit differently than the archives you normally provide. The `opensearch` package is minimal by default, without any plugins. Those are available as [additional packages](https://archlinux.org/packages/?q=opensearch+plugin). Therefore, I would provide the installation instructions like this:

opensearch example

Same goes for OpenSearch Dashboards:

opensearch-dashboards example

 If you think this doesn't suit you we can provide the Arch Linux platform on the "Minimal Distributions" list only.

### DIFF
--- a/_artifacts/_README.markdown
+++ b/_artifacts/_README.markdown
@@ -16,7 +16,7 @@ role: cli
 artifact_id: opensearch-cli 
 # `version` is the semver version
 version: 1.0.0
-# `platform` is the identifier for the OS or platform. Should be `linux`, `windows`,`macos`,`freebsd`
+# `platform` is the identifier for the OS or platform. Should be `linux`, `windows`,`macos`,`freebsd`,`archlinux`
 platform: linux
 # `architecture` should be `arm64`,`x86`,`x64`
 architecture: arm64
@@ -31,8 +31,9 @@ type: zip
 # `signature` is the download link to the signature file
 signature: https://artifacts.opensearch.org/opensearch-clients/opensearch-cli/opensearch-cli-1.0.0-linux-arm64.zip.sig
 # others:
-# `link` used for `docker_hub` and `freebsd_package_name`, the page that has information non-artifact
+# `link` used for `docker_hub`, `freebsd_package_name` and `archlinux_package_name`, the page that has information non-artifact
 # `freebsd_package_name` the `pkg` name, used in the template only for freebsd
+# `archlinux_package_name` the `pkg` name, used in the template only for Arch Linux
 
 ---
 ```
--- a/_artifacts/opensearch-1.2/x64/opensearch-1.2.4-archlinux-x64.markdown
+++ b/_artifacts/opensearch-1.2/x64/opensearch-1.2.4-archlinux-x64.markdown
@@ -1,0 +1,12 @@
+---
+role: daemon
+artifact_id: opensearch
+version: 1.2.4
+platform: archlinux
+architecture: x64
+slug: opensearch-1.2.4-archlinux-x86_64
+category: opensearch
+type: system-package
+archlinux_package_name: opensearch
+link: https://archlinux.org/packages/community/x86_64/opensearch/
+---
--- a/_artifacts/opensearch-1.2/x64/opensearch-1.2.4-min-archlinux-x64.markdown
+++ b/_artifacts/opensearch-1.2/x64/opensearch-1.2.4-min-archlinux-x64.markdown
@@ -1,0 +1,13 @@
+---
+role: daemon
+artifact_id: opensearch-min
+version: 1.2.4
+platform: archlinux
+architecture: x64
+slug: opensearch-1.2.4-min-archlinux-x86_64
+category: opensearch
+type: system-package
+archlinux_package_name: opensearch
+link: https://archlinux.org/packages/community/x86_64/opensearch/
+security_warning: true
+---
--- a/_artifacts/opensearch-1.2/x64/opensearch-dashboards-1.2.0-archlinux-x64.markdown
+++ b/_artifacts/opensearch-1.2/x64/opensearch-dashboards-1.2.0-archlinux-x64.markdown
@@ -1,0 +1,13 @@
+---
+role: ui
+title: OpenSearch Dashboards
+artifact_id: opensearch-dashboards
+architecture: x64
+platform: archlinux
+type: system-package
+category: opensearch-dashboards
+slug: opensearch-dashboards-1.2.0-archlinux-x64
+version: 1.2.0
+archlinux_package_name: opensearch-dashboards
+link: https://archlinux.org/packages/community/x86_64/opensearch-dashboards/
+---
--- a/_artifacts/opensearch-1.2/x64/opensearch-dashboards-1.2.0-min-archlinux-x64.markdown
+++ b/_artifacts/opensearch-1.2/x64/opensearch-dashboards-1.2.0-min-archlinux-x64.markdown
@@ -1,0 +1,14 @@
+---
+role: ui
+title: OpenSearch Dashboards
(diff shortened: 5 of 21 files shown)


inglor:
I don't this older versions should be included here. Only the version in the official repositories (the latest) is the one Arch Linux will support.

hashworks:
That sounds reasonable. I guess every merge request should remove the old (now unsupported) version. Since we then provide only one version we should also link to the [package pages](https://archlinux.org/packages/community/x86_64/opensearch/).

stockholmux:
@hashworks This amazing! I'm working on some refactoring of the downloads page so I'm holding on merging this one for a little bit (merging makes it go live in the deployment pipeline).

nateynateynate:
Hi @hashworks - first, I wanted to thank you for your contribution! I was quite pleased to see this PR come in. I was wondering if you could help me here. 

I'm going through an install on my own Archlinux box just for a sanity check, and I didn't have any luck getting dashboards to run (it installed fine). I kept getting this error: 

image

When clicking through the dependencies at https://archlinux.org/packages/community/x86_64/opensearch-dashboards/, I noticed that it has a requirement for https://archlinux.org/packages/community/x86_64/nodejs-lts-erbium/ which is at 12.22.9, but OpenSearch Dashboards requires 12.22.8 to run. 

Are you seeing this same issue at all?

hashworks:
I'm sorry, this must have slipped by me. Please try out the new `1.2.0-2`.

Since we only package the current LTS versions of node (thus, minimum 12.22.9) we don't provide the [currently listed 10.24.1](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/1.2/package.json#L471). Since in my tests v12 worked just fine and there are plans to move to v14 (#920) I replaced the version requirement in the `package.json` with our current `nodejs-lts-erbium` version. This works fine for builds, but may fail in runtime since the `nodejs-lts-erbium` patch version (8 -> 9) might change and `node_version_validator.js` checks for the exact version.

For now, I've removed `require('./node_version_validator');` (and thus the runtime check). This shouldn't be required once v14 is used, but I would still advise to only check major and minor versions, not patch versions, like:
```node
var isVersionValid = requiredVersion.match(/^v(\d+\.\d+)/)[1] === currentVersion.match(/^v(\d+\.\d+)/)[1]
```
But let's discuss that in another issue.

Do you want people to be notified about the adjusted NodeJS version on the download page?

nateynateynate:
Thanks for the update @hashworks.

I think we might have an opportunity for collaboration here. 

We'd really love to see the post-install experience to match what is currently distributed by the project, but I can't help but notice a bit of divergence. 

It might be good to explore together how we can fill in the gaps for the Arch users so they can get up and running in a way that matches the OpenSearch documentation more closely in the areas around security and configuration. Glad to work with you on this or hear your thoughts on what would be idiomatic for Arch users.

hashworks:
> I can't help but notice a bit of divergence

If you list the divergences you noticed I would happily address them or provide reasoning for the adjustments :)

> Glad to work with you on this or hear your thoughts on what would be idiomatic for Arch users.

I guess the basic idiom is simplicity. Arch users expect the installed packages to be as close to upstream as possible without unnecessary additions or modifications, including configuration files. All adjustments should be distro related, not something the packager "prefers". We also want to ensure flexibility – if users decide that they don't need a security plugin they should be able to do so, which is why the plugins are provided as separate packages.

Therefore, the [upstream configuration file](https://github.com/opensearch-project/OpenSearch/blob/1.2.4/distribution/src/config/opensearch.yml) is used without any modifications. For opensearch-dashboards a single adjustment is made since we expect users to run it using the provided systemd service.

```bash
sed -r 's|#(logging.quiet:) false|\1 true|' -i config/opensearch_dashboards.yml
```

stockholmux:
@hashworks OK. I get where you're coming from with regards to packaging & thanks for explaining the Arch philosophy. 

I think the rub is with the packaging - individual plugins vs together as a single unit. Currently, OpenSearch [does distribute a `-min` package but with a serious warning before download](https://opensearch.org/artifacts/opensearch/opensearch-1-2-4-min-linux-x64.html). And the plugins aren't distributed individually on the downloads page. This is intentional. From the project perspective, the way this is split up is largely historical and will likely change radically at version 2 and through version 3. So, that leads Arch users to be put at some pretty serious disadvantages:

1) Most of the documentation will not apply to those who install OpenSearch through Arch. The documentation assumes SSL and authentication, as an example.
2) As the project changes and the constituent units (plugins), which are now represented as separate packages on Arch, will go away or change function, so if someone is using Arch and expecting their install scripts to behave the same way in, say, 6-8 months, they're going to be pretty disappointed.

Point 1 is easier to bridge. One thing that comes to mind is making the downloadable artifact on the website a script that uses `pacman` to install and setup an equivalent experience to what someone would get if they installed from the .tar.gz or using docker. Someone could independently install OpenSearch with pacman but at least if they get something from the opensearch.org download page, it will be like-for-like. This is something I'm sure I can find some folks to collaborate on.

Point 2 is tougher. While breaking changes happen all the time across a distro - I'd anticipate the combination of how OpenSearch is broken up now vs where it's heading causing more headaches that love for arch users. I'd be very interested how there could be collaboration in making this a smooth transition for everyone.


Thoughts on any of this?

dblock:
@peternied WDYT ^?

peternied:
First off - this is amazingly cool!

> I wonder if it would make sense for the security plugin to just disable TLS when it isn't configured and provide a warning instead.

I'm not sure if disabling the feature is correct vs making TLS configuration have better defaults; however, this is a great issue for us to tackle.  I've created https://github.com/opensearch-project/security/issues/1631

> One way to do this would be a meta package

I like this idea, a logical framework that can be consistent even if we reshape the 'physical' layout of plugins.

dblock:
Can this be merged (soon)? We're talking about Arch Linux in https://github.com/opensearch-project/project-website/pull/670

hashworks:
> One way to do this would be a meta package

Since [#1631](https://github.com/opensearch-project/security/issues/1631) has been closed, I won't add such a drastic change (which might break current installations). Any other way we could merge this? Adjustments to the text, additional warnings, wiki-notes, …

dblock (dismissed):

hashworks:
What should we do with this PR? If this will never be merged I can stop adding the new versions.

hashworks:
> OK if I read through and work to get back to you soon?

Sure!

bbarani:
Anyone looking in to this PR?  @krisfreedain @stockholmux

nateynateynate:
Can I get an assist on whomever is involved here on what the next steps are? @dblock, @peternied, @hashworks, @bbarani ? 

It would seem that Arch Linux has some requirements for their packages being as minimal as possible. I'd sure like to get this one merged some day.

dblock:
Apologies for the ... delay. Looks like archlinux artifacts continue being released (awesome!) but we are still failing to publish links to the on the website. @krisfreedain could you please take a look?

[Catch All Triage - Attendees [1](https://github.com/dblock/), [2](https://github.com/krisfreedain), [3](https://github.com/andrross), [4](https://github.com/gkamat), [5](https://github.com/getsaurabh02)]

krisfreedain:
@nateynateynate @kaimmej - can we look at this one this week? I think we're at a place where this distro needs to be included in the download each time - looks like the team is doing a build each time - thanks

nateynateynate:
https://github.com/opensearch-project/project-website/pull/3107

nateynateynate:
Ok it looks like you're live from #3117 Enjoy! :-)

Issue #661, opened 2022-01-18
Publish an End of of Life / LTS / Supported Versions Policy 


julian-alarcon:
Currently there is no place to look for the supported versions by OpenSearch, or is currently just the latest release supported?
Old project has a good and clear reference of supported versions and a specific plan https://www.elastic.co/support/eol

**Describe the solution you'd like**
There should be a clear document mentioning the current supported versions, if there is any rule to support latest 2-n releases and even if there is any LTS kind of support for older versions (E.g. 1.0)

stockholmux:
I recall @jcgraybill was working on something like this.

jcgraybill:
Yes, the current policy is straightforward - all maintenance and development work is performed against the latest version (which today is OpenSearch 1.2.4 and OpenSearch Dashboards 1.2.0). The project follows follows [the semantic versioning specification](https://opensearch.org/blog/technical-post/2021/08/what-is-semver/) (https://opensearch.org/blog/technical-post/2021/08/what-is-semver/) for assigning version numbers to releases, so you should be able to upgrade to the latest version of any software without encountering incompatible changes.

The question becomes more interesting when we release OpenSearch 2.0.0, since that version very likely will have some number of incompatible changes. At that point, we will continue to perform security and bug-fix maintenance on the latest 1.y.z version of OpenSearch for a period of time. 

We're in the process right now of figuring out what that maintenance window will be. A big input to that is the EOL, support, and maintenance plans of the project's dependencies, of which there are many. I'd like to be able to sign up for a concrete maintenance window for OpenSearch 1.y.z before settling on a launch schedule for OpenSearch 2.0.0. Once we do so, we will indeed post this to the website.

stockholmux:
I tend to agree with @dblock here and might go a step further. It would be good to have a simple URL (e.g. opensearch.org/supported-versions) to point to that tells the simplest story:
- semantic versioning
- currently supported version
- where feature are being added

For example, I always find myself looking at the [Node.js releases page](https://nodejs.org/en/about/releases/). Node.js has a moderately complicated story here but their page is clear and concise.

(it might also be a good place to eventually start talking about how the OpenSearch project deals with dependencies which is enlightening and, I think,  poorly understood)

dblock:
We now have an FAQ, https://opensearch.org/faq#q3.28, but we still need a support page that explains which versions are _currently_ supported and what their EOL is. I'm reopening this.

PR #693, opened 2022-02-15
Add new FAQ covering versioning and maintenance approach

jcgraybill:
### Description
Add new FAQ covering versioning and maintenance approach. Also renaming two images to shorten file names to be NTFS compatible.
 
### Issues Resolved
https://github.com/opensearch-project/project-website/issues/661

### Check List


_faqs/03-28-which-version-of-opensearch-should-i-use.markdown
```diff
@@ -0,0 +1,10 @@
+---
+question: Which version of OpenSearch should I use?
+category: Upgrading to OpenSearch
+---
+
+or OpenSearch and other software in the OpenSearch project, new features and active development always takes place against the newest version. The OpenSearch project follows the [semantic versioning specification](https://opensearch.org/blog/technical-post/2021/08/what-is-semver/) for assigning version numbers to releases, so in most cases you should be able to upgrade to the latest version of any software without encountering incompatible changes.
```
elfisher, line 6: I think there is a missing "F" at the start of this paragraph

elfisher (changes requested):
LGTM one minor suggestion.

_faqs/03-28-which-version-of-opensearch-should-i-use.markdown
jcgraybill: 🤦‍♂️ Fixed, thanks.

elfisher (approved):
LGTM!

stockholmux (approved):
LGTM

PR #757, opened 2022-04-07
- added maxcluster logo & link to _partners

dmonkey4224:
### Description
Adds maxcluster as a partner (after communication with Joshua Bright)
 
### Issues Resolved
-


dmonkey4224:
I amended the commit and added the DCO as the guide stated it as the first fix option . I hope that works for you.

stockholmux:
We still have an unhappy DCO bot :(

So, the complexity here is that you had contributes the changes - the DCO will only work when _all_ commits are signed (even the ones in the past (e.g. https://github.com/opensearch-project/project-website/pull/757/commits/2cd86cf200f65415c2115ff99fdf065ee02227cd). I would suggest a [rebase as per the bot's instructions](https://github.com/opensearch-project/project-website/pull/757/checks?check_run_id=5880791437)).

The option would be to start over and re-contribute (see [Option 4](https://github.com/src-d/guide/blob/9171d013c648236c39faabcad8598be3c0cf8f56/developer-community/fix-DCO.md#fix-option-4-almost-avoid-git))

PR #849, opened 2022-05-31
Add additional libs for jekyll

willyborankin:
Added 2 additional libs which jekyll needs to start on ruby 3.x
- webrick
- xmlrpc

### Description
[Describe what this change achieves]
 
### Issues Resolved

### Check List

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,6 +8,8 @@ source "https://rubygems.org"
 # This will help ensure the proper Jekyll version is running.
 # Happy Jekylling!
 gem "jekyll", "~> 4.2.0"
+gem "webrick", "~> 1.7.0"
+gem "xmlrpc", "~> 0.3.2"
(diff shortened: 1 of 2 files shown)


dblock:
I fixed the same problem in #1456. Can we please get this and some of the other old PRs merged?

ananzh (approved):
Don't know Ruby very well. My understanding is that Jekyll uses the webrick gem as a built-in server for development purposes. Starting from Ruby 3.0, webrick has been removed from the standard library, which is why we see a "cannot load such file" error.

This fix is to add the webrick gem to the Gemfile to ensure that it is installed and available when running the Jekyll server. I think the adding xmlrpc gem is not related to the error. Maybe it needs to be added by other Jekyll plugins or themes.

dblock:
@ananzh indeed, can we merge either PR and close the other?

krisfreedain:
Already taken care of in above mentioned PR. thank you

PR #931, opened 2022-07-28
update to maintenance policy

CEHENKLE:
Signed-off-by: CEHENKLE <henkle@amazon.com>

### Description
Updated to clarify that folks can open PRs against unsupported versions, but we will not generate new releases.
 

### Check List

### DIFF
--- a/releases.md
+++ b/releases.md
@@ -47,6 +47,9 @@ Sometimes an incompatible change is unavoidable. When this happens, the software
 
 The duration of the maintenance window will vary from product to product and release to release. **By default, versions will remain under maintenance until the next major version enters maintenance, or 1 year passes, whichever is longer.** Therefore, at any given time, the current major version and previous major version will both be supported, as well as older major versions that have been in maintenance for less than 12 months. Please note that, maintenance windows are influenced by the support schedules for dependencies the software includes, community input, the scope of the changes introduced by the new version, and estimates for the effort required to continue maintenance of the previous version. 
 
+While the software maintainers will not automatically back-port fixes or features to versions outside of the maintenance window, PRs with back-ports to unsupported versions are welcome and review of them will follow the project’s [review process](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#review-process); however, the software maintainers may not back-port fixes or features to versions outside of the maintenance window and no new releases will be created for accepted PRs. That said, interested parties are welcome to [create their own distribution](https://github.com/opensearch-project/opensearch-build#building-and-testing-an-opensearch-distribution) from the updated source after the PRs are merged.
+
+
 <div class="table-styler"></div>
 
 | Major Version | Latest Minor Version |   Status    | Initial Release | Maintenance Window Start | Maintenance Window End |


releases.md
```diff
@@ -47,6 +47,9 @@ Sometimes an incompatible change is unavoidable. When this happens, the software
...
 
+While the software maintainers will not automatically back-port fixes or features to versions outside of the maintenance window, PRs with back-ports to unsupported versions are welcome and review of them will follow the project’s [review process](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#review-process); however, the software maintainers may not back-port fixes or features to versions outside of the maintenance window and no new releases will be created for accepted PRs. That said, interested parties are welcome to [create their own distribution](https://github.com/opensearch-project/opensearch-build#building-and-testing-an-opensearch-distribution) from the updated source after the PRs are merged.
```
pjfitzgibbons, line 50: I'm slightly confused by the distinction between "backports to unsupported versions are welcome" and "software maintainers may not back-port fixes... to versions outside maintenance window" ?

Maybe I'm anticipating a sentence between about how the regular maintenance cycle will include backported fixes within the regular maintenance window?

I feel like this paragraph (L50) has an implicit context of "PR to apply to non-maintained versions", and that my understanding of the outcome is that a PR may be accepted and merged into unsupported version by maintainers, but no automated build nor release will be created by those merges.

CEHENKLE: So I think there are two things going on:  One is the review process to accept a PR (which is true of all PRs, backports, features and bugs), the other is how we handle releasing artifacts.

For reviewing PRs,  we know that opening a PR is a significant effort for folks, but sometimes the maintainers cannot accept them.  That can be because they don't align with the long term plans for the project, or because they're not well tested, or anything in between.  Opening a PR is not a guarantee it'll get merged in. 

The second part is around when we release artifacts.  The process for a supported version is that someone raises a  PR.   it's reviewed and if it's accepted, it's merged in to the main branch.  From there maintainers can chose to open back ports to the supported versions (2.x and 1.3) and then it'll get released as an artifact on the next scheduled release.  

But for fixes to non-supported versions, the maintainers will only review backported PRs, if someone explicitly raises them.  If the PR is accepted, they'll merge it to the targeted version(s).  But then we *won't* release a new artifact of the unsupported software, however folks are welcome to build their own. 

More information about:

- The PR review process:  take a look at https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#review-process.  
- The release train for supported versions:  https://opensearch.org/releases.html
- OpenSearch branching and backporting strategy:    https://github.com/opensearch-project/.github/blob/main/RELEASING.md#branching

Make sense?

joshuarrrr: I think 

> But for fixes to non-supported versions, the maintainers will only review backported PRs, if someone explicitly raises them. If the PR is accepted, they'll merge it to the targeted version(s). But then we won't release a new artifact of the unsupported software, however folks are welcome to build their own.

is more clear than what's currently in the text. One of the issues is "PRs with back-ports to unsupported versions" -  I get now that it means PRs that target a non-supported release branch instead of `main`, but I didn't really grok that on the first read. For "from the updated source after the PRs are merged", it might help to say "from the updated branch after the PRs are merged" 🤷

krisfreedain:
@CEHENKLE @pjfitzgibbons @joshuarrrr are we good to go on this one?

CEHENKLE:
Dave stole a march on me with https://github.com/opensearch-project/project-website/pull/1062/files

PR #1069, opened 2022-10-10
Hacktoberfest blog post: don't use `git commit -m`

rursprung:
### Description
Don't suggest the usage of using the `-m` option to define a commit-msg directly when running the command. This leads to super ugly commit messages as you can essentially only define a single line and don't have any guidance on the length. As Hacktoberfest is esp. used by new developers this will lead them down the wrong path.

Probably the best explanation of proper commit messages can be found in [this blog post from 2008][tbaggery] (Linus Torvalds - the guy who invented both Linux and Git - also [confirmed][linus] that this lists all the rules for good commit messages).

[tbaggery]: https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html
[linus]: https://github.com/torvalds/linux/pull/17#issuecomment-5663733

Signed-off-by: Ralph Ursprung <Ralph.Ursprung@avaloq.com>

### Issues Resolved
n/a

### Check List

### DIFF
--- a/_posts/2022-10-03-hacktoberfest-2022.md
+++ b/_posts/2022-10-03-hacktoberfest-2022.md
@@ -46,7 +46,7 @@ I am so glad you asked! For contributions, I recommend you start small. OpenSear
 
 Once you have found an issue you would like to work on, you should check out their `CONTRIBUTING.md` file. It will take you through the process of contributing to that particular repo. Also, if you are contributing to a particular issue, it is helpful to comment on it to let others know you are working on it. 
 
-Generally, the process for contributing involves forking the repository, making your edit (don’t forget to sign your commits using -s in your commit message: `git commit -s -m "<your commit message>"`), and then submitting a pull request (PR). This guide goes into much more detail: [First Contribution Guide](https://github.com/firstcontributions/first-contributions).
+Generally, the process for contributing involves forking the repository, making your edit (don’t forget to sign your commits using -s in your commit message: `git commit -s`), and then submitting a pull request (PR). This guide goes into much more detail: [First Contribution Guide](https://github.com/firstcontributions/first-contributions).
 
 If you have questions along the way, you can always @ mention one of the maintainers on the issue. You can find the maintainers in the `MAINTAINERS.md` file for that repo. Please be patient as the maintainers may take several days to reply depending on their capacity. If more than a few days goes by without a response, feel free to reach out to another maintainer. 
 


rursprung:
CC @dtaivpp & @krisfreedain as this is a follow-up to your PRs

dtaivpp:
Actually @rurspring I am going to disagree here. While it might not be the best way to add commit messages it is the most straightforward for new people. Vi can be challenging and a major hurdle to new engineers. 

What I would accept is if you wanted to add a short section that falls underneath that line about a better way to commit that talks through using vi and best practices I'd happily review it.

rursprung:
> While it might not be the best way to add commit messages it is the most straightforward for new people. Vi can be challenging and a major hurdle to new engineers.

let's be honest: 90% of the people will anyway use a graphical Git tool and never interact with the console. and those using the console will know what they're doing (or learn it very fast 😈). i'm just trying to avoid people copy & pasting `-m` and thinking it's a good idea.

you're also not providing any commands on how to clone, create a branch, etc. in the blog post - so if somebody is using a console they'll anyway have to have a basic understanding of the git tooling (or a good guide on the side).

maybe it makes more sense to just explicitly write that you have to add a line along the pattern of `Signed-off-by: Foo Bar <foo.bar@example.org>` (all `-s` does is adding that automatically for you based on your git-config settings, you can still edit it at will if you want).


PS: i will probably (eventually, when i get around to it... maybe [soon](https://wowwiki-archive.fandom.com/wiki/Soon)) create a forum post to address the general commit history quality in the opensearch project, in the hope of getting it improved. i've seen a lot of commits with very unhelpful commit messages and think that awareness here is lacking in general 😞

dtaivpp:
So I agree many might use a graphical git tool. If you would like to add an excerpt about a better way beneath that paragraph that is okay. Commit signing is one of the biggest issues we have with contributions currently and commit -s is the most straightforward way for us to ensure that the signature matches the commit. Same goes with messages. -m gets them going with the least overhead. 

Also, we do provide a guide on creating a Fork and PR. It's the last line of the paragraph you reference. 
https://github.com/opensearch-project/project-website/blob/9f61f381368fc64132ff99ed56d4b676ab054634/_posts/2022-10-03-hacktoberfest-2022.md?plain=1#L49

rursprung:
> If you would like to add an excerpt about a better way beneath that paragraph that is okay.

hm, i think then we can also leave it be. once you give somebody a gun to shoot themselves (and you) with they're usually not going to relinquish it again any time soon. because adding "or leave the `-m` away and use an editor to write the commit message to provide more details and format your message nicely" probably sounds like "or do more work for which you don't see any value right now" to most people reading it?)
i did indeed miss the link you had underneath, and it also suggests using `-m` 😢.

i guess if/when we get to the discussion about commit messages in the opensearch projects we might then also get back around to this (or just say "let them start however they want but before we'll merge it they'll have to do some re-wording if it's not ok").

so if you agree we can also close this PR

PR #1086, opened 2022-10-18
adding 2023 proposed schedule

CEHENKLE:
Signed-off-by: CEHENKLE <henkle@amazon.com>

### Description
Adds in proposed 2023 schedule
 
### Issues Resolved
We don't stop releasing at the end of 2022 ;) 

### Check List

### DIFF
--- a/releases.md
+++ b/releases.md
@@ -5,7 +5,7 @@ title: Release Schedule and Maintenance Policy
 ---
 
 
-_Updated October 6, 2022_
+_Updated October 20, 2022_
 
 [Release Schedule](#release-schedule) &middot; [Maintenance Policy](#maintenance-policy)
 
@@ -15,6 +15,8 @@ _Updated October 6, 2022_
 ## Release Schedule ##
 
 For more information on the changes planned for each release, please see the [Project Roadmap](https://github.com/orgs/opensearch-project/projects/1)
+
+### 2022 ###
 
 <div class="table-styler"></div>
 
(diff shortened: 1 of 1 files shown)


releases.md
```diff
@@ -29,7 +31,31 @@ For more information on the changes planned for each release, please see the [Pr
 | 1.3.6          | September 30, 2022 | October 6, 2022    |
 | 2.4            | November 3, 2022   | November 10, 2022  |
 | 1.3.7          | December 1, 2022   | December 8, 2022   |
-| 2.5            | January 10, 2023   | January 12, 2023   |
+| 2.5            | January 10, 2023   | January 17, 2023   |
```
seanneumann, line 34: You can just remove this since its moved to the 2023 section.

releases.md
```diff
@@ -29,7 +31,31 @@ For more information on the changes planned for each release, please see the [Pr
...
+
+### Proposed 2023 schedule ###
+
+
+
+Note:  We have not added a major release to the 2023 schedule yet.  If/when we add one, it will replace a minor release in the 2.x line.  See below for criteria for a major releases.
+
+| Release Number| Code Freeze Date  | Release Date          |
```
seanneumann, line 42: Some of these don't follow the six week model. Some are seven. By design?

releases.md
```diff
@@ -29,7 +31,31 @@ For more information on the changes planned for each release, please see the [Pr
...
+| 2.7.0         | April 11th        | April 18th            |
+| 1.3.10        | May 11th          | May 18th              |
+| 2.8.0         | May 30th          | June 6th              |
+| 1.3.11        | June 22nd         | June 29th             |
+| 2.9.0         | July 11th         | July 18th             |
+| 1.3.12        | August 3rd        | August 10th           |
+| 2.10.0        | August 22nd       | August 29th           |     
+| 1.3.13        | September 14th    | September 21st        |
+| 2.11.0        | October 10th      | October 17th          |
+| 1.3.14        | November 9th      | November 16th         |
```
peternied, line 57: We have gotten much better at preparing and releasing patches, what do you think about betting that we can compress the 7 day time to something smaller?  Being bold, what do you think about targeting releasing at the end of the day of code freeze?

CEHENKLE: I asked @bbarani about this, and he asked to keep the 7 days.  are you really, really sure @bbarani we need to freeze for 7 days?

releases.md
CEHENKLE: In part.  I was dodging around major US holidays (my bias).   We do say "approximately" every 6 weeks :)  

If there are any you think we should tighten up to 6 weeks, LMK.

releases.md
CEHENKLE: Removed

nateynateynate (approved):
Looks good to me. I'd merge if I could. ;-)

CEHENKLE:
> If this is something we can accommodate (given that it might result in additional releases), I would consider releasing the 1.x and 2.x lines together using the same cadence to account for the fact that sometimes we need to address security issues that affect both, and having the fix in just one but not the other is not the best.
> 
> This is especially problematic when those fixes are for security issues we are disclosing ourselves via advisories, as we would be making the issue public but not providing a way for users of 1.x or 2.x to address them.

We are already committed to doing ASAP patches for critical issues, and I want to doublecheck (@bbarani) but I think we release them in both supported lines (1.x and 2.x) at the same time.  If we don't we should :) 

For highs and below, it looks like we're coalescing around 60 days from notification, and since there's 3-4 weeks between any 2.x and 1.x release, I think we're still good?  

The bottle neck for me is testing, since there's still some manual testing that needs to happen.  I think we should lay the calendar out assuming they're separate, but I see the value in being able to release them both simultaneously.  @bbarani what do you think about taking that as a goal (to do simultaneous releases ever 4 weeks) for next year? What would we have to tighten up/automate to make that happen?

CEHENKLE:
@opensearch-project/project-website Hey folks, can I get a review?  Thanks!

peternied (approved):

releases.md
seanneumann: Nice.

seanneumann (approved):

elfisher (approved):
LGTM!

PR #1218, opened 2022-12-02
#1115 Momentum Stats Carousel

### LINKED ISSUE #1115: [FEATURE] Momentum stats carousel.
**Is your feature request related to a problem? Please describe.**

In September at OpenSearchCon, we shared a number of impressive metrics around the project. We need to highlight these publicly on the website. Evaluate proper placement; idea - near top of right-hand side bar- Evaluate responsibility and cadence for updating (once a month? once a quarter?)

**Describe the solution you'd like**

Some kind of slideshow / accordion / carousel, somewhere near the top of the right hand column. Above or below the Download buttons? Fade out old / fade in new at regular intervals. 

Make it look cool. I vote skulls and fire and laser beams, but that's probably outside our branding guidelines.  Presentation layer will require some debate. 

**Describe alternatives you've considered**

Adding an image and changing it regularly via PR. 

**Additional context**

The process of updating and adding new 'stats' should be considered, as well as the potential of numbering them by hand. Allow for an _include keyword to be passed that tells it to show in random order and how long to wait until the image / div flips.

nateynateynate:
Signed-off-by: Nate Boot <nateboot@amazon.com>

### Description
Adds a section currently above the feedback section that shows 'momentum stats'. 

Stats take the form of markdown and yaml frontmatter, and all live inside of _momentumstats. They can be an `html` stat, or an `image` stat. Image stats just show the image and links it to `stat_url` frontmatter value. (See `template.markdown` in _momentumstats)

Their presence is determined by the frontmatter attribute `display` being `true`. 

If there are no stats with a display value of true, the section does not appear. 
 
### Issues Resolved
#1115 

### Check List

### DIFF
--- a/_includes/momentumstats-carousel.html
+++ b/_includes/momentumstats-carousel.html
@@ -1,0 +1,46 @@
+{% assign mstats = site.momentumstats | where: "display", "true" %}
+
+{% if mstats.size > 0 %}
+	<style type="text/css">
+		img.momentumstat-image
+		{
+			width: 420px;
+		}
+		
+	</style>
+
+	<h3>
+		Check out some of our momentum!
+	</h3>
+
+	<div class="momentumstats-carousel-container">
+
+		{% for momentumstat in mstats %}
+			<div class="momentumstats-slide-container" id="momentumslide{{ forloop.index0 }}"
+				{% if forloop.first %}
+					style="display: inline;"
+				{% else %}
+					style="display: none;"
+				{% endif %}
+				>
+				{% include momentumstats-carousel-slide.html momentumstat=momentumstat %}
+			</div>
+		{% endfor %}
+	</div>
+
+	<script language='javascript'>
+		setInterval(function() {
+			thingies = jQuery(".momentumstats-slide-container");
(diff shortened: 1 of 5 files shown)


_includes/momentumstats-carousel.html
```diff
@@ -0,0 +1,46 @@
...
+				{% endif %}
+				>
+				{% include momentumstats-carousel-slide.html momentumstat=momentumstat %}
+			</div>
+		{% endfor %}
+	</div>
+
+	<script language='javascript'>
+		setInterval(function() {
+			thingies = jQuery(".momentumstats-slide-container");
```
astephanus, line 33: Written as it is  here the `thingies` symbol will be attached implicitly as a property of `window`. Is this the intention, or was omitting `const`, `var`, or `let` accidental? Note that if an eventual change was put in place that put global scope into Strict Mode that this would throw a `TypeError` exception.

nateynateynate: Gotcha. Nope, this was me sacrificing quality for something that would function. I can refactor to avoid the TypeError.

PR #1381, opened 2023-02-27
Request for Smart Search Tools and Weblink Technology as partners

sribalajivelan-weblinktechs:
### Description
[Describe what this change achieves]
 
### Issues Resolved
[List any issues this PR will resolve]

### Check List
- [x] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the BSD-3-Clause License.

PR #1470, opened 2023-03-27
Include link checker in CI

dblock:
### Description

Similar to https://github.com/opensearch-project/documentation-website/pull/3584.

- Run link checker in CI, fail PRs on broken links.
- Improve the homegrown link-checker a bit, make HEAD requests, etc.
- Fix many broken links.
- Closes https://github.com/opensearch-project/documentation-website/issues/3583
 
### Check List


dblock:
Thanks for the review @krisfreedain!

1. Yes, the build will fail if a link is broken. I will move this into a cron, but in the past I've seen that cause us to stop paying attention to broken links. Are you sure? Any contributor that runs into a broken link will tend to fix it if they run into that obstacle, which I think accomplishes the best outcome of links being rarely broken.
2. I think that partner no longer exists, the website is MIA. Let me know what you want me to do with it: I could instead remove the link, or something else?

krisfreedain:
1. Yeah, I'd love to have a weekly / monthly report on broken links if you build it out as a cron job. That would be helpful (as demonstrated by the updates you found) - but I don't want to force-break the build. When a contributor files a PR, one of us that are active maintainers should be reviewing them prior to publish. Broken links are less common at that time - but a report to catch them if they break later is actionable. 
2. Yeah - let's keep it, but I'm OK with removing the link - myself and @pajuric can follow up with them from the partner side and see what is going on. (this partner was added via PR: https://github.com/opensearch-project/project-website/pull/1381 ) 

thanks @dblock !

krisfreedain (approved):
this is great - thank you @dblock

PR #1488, opened 2023-04-01
Corrects partners collection URL configuration

astephanus:
### Description
Corrects partner page collection URL configuration by including the same configuration in `_config.yml` into `_ config-dev.yml`.

### Check List


astephanus:
This was not the correct resolution to the problem of broken partner page URLs. Closing PR.

PR #1492, opened 2023-04-02
Faster link checker using ruby-link-checker.

dblock:
### Description

I extracted an external link checker into https://github.com/dblock/ruby-link-checker. Does about 50 URLs/second on my M1. Makes it a lot simpler here, notably no need to deal with retries, error codes, etc. The remaining code is for local links, which makes it difficult to use an external link checker like lychee.
 
Successful run in https://github.com/dblock/project-website/actions/runs/4598091548/jobs/8121654192. Took about half the time from the sequential checker. It could go a lot faster with higher concurrency, but some servers 429 too often, which could be handled better (https://github.com/dblock/ruby-link-checker/issues/2).

### Check List


dblock:
Updated to run on a daily cron and on demand.

dtaivpp (approved):
Looks great to me

PR #1616, opened 2023-05-22
Fix next Community Meeting Time Zone

smortex:
The event on https://opensearch.org/events/2023-0523/ currently says:

Date: Tue, May 23, 2023 Time: 15:00 (UTC -8)

Translated to my Time Zone (UTC-10), this is:

> Date: Tue, May 23, 2023
> Time: 13:00 (UTC -10)

But the meetup page says:

Tuesday, May 23, 2023 at 12:00 PM to Tuesday, May 23, 2023 at 1:00 PM UTC−10

So one hour offset: from 12:00 to 13:00 on meetup and 13:00 to 14:00 on the website.

Assuming the time-zone is in fact "PDT" which is "UTC-7" or "UTC-8"
depending on Daylight Saving Time, this should be set to "UTC-7".

### DIFF
--- a/_events/2023-0523.markdown
+++ b/_events/2023-0523.markdown
@@ -1,7 +1,7 @@
 ---
 
 eventdate: 2023-05-23T15:00
-tz: UTC -8
+tz: UTC -7
 title: OpenSearch Community Meeting - 2023-05-23
 online: true
 signup:


dtaivpp:
This is a great catch! I will update that and we will go through and update some of the others.

dtaivpp (approved):
Looks good!

PR #1636, opened 2023-05-30
Reduce javascript components

astephanus:
### Description
Refactors some custom elements from the extended partner pages from their implementation as Custom HTML Elements with Shadow DOM into Jekyll Liquid templates with global SCSS using BEM class name notation. Business logic that had been performed by JavaScript in the custom elements has been refactored into Liquid logic, or driven by changes to the structure of the Front Matter for the partner's to eliminate the need for any business logic in favor of minimizing any domain model logic in the view.

### Check List


dtaivpp:
Hey @astephanus how are you testing for the social icons? I was trying to find a partner with them but couldn't find one.

astephanus:
@dtaivpp Yes. There are very few partners whom have provided social links. The list is as follows:

- CERN
- Searchuim AI
- Sematext
- Tracetest
Thankfully in double checking I did happen to notice that some styling that had been in the previous custom element shadow DOM implementation did not get coverted correctly for those icons. However, that did not affect whether or not they were displayed.

astephanus:
> Ah thank you! Idk how I missed them. Will review and get back to you shortly!

@dtaivpp An easy way to check to see which partners have social links is to run this in the terminal from the project directory: `grep social_links ./_partners/*.md`. That will give you a list.

dtaivpp:
Okay looking this over and I am a little wary about the spacing of the icons. Despite them being evenly spaced and square they don't look it. Maybe is there a square set of icons we could use instead?

Sematext
Cern

PR #1640, opened 2023-06-01
Localize the date and time of events

smortex:
This is a follow-up to #1616.

Use progressive enhancement to show the user the date and time of events in their time zone:

image

For on-site events, if the user is located in a "foreign" time zone, display both the date and time of the event in the local time zone of the user and the local time zone of the event:

image

			     
If JavaScript is disabled, a single timestamp in RFC822 format is displayed.

Screenshots are from a browser on a system configured to speak French.

### DIFF
--- a/_includes/event.html
+++ b/_includes/event.html
@@ -2,7 +2,7 @@
   <a href="{{ event.url }}">{{event.title}}</a>
   <p> {{ event.content | strip_html | truncatewords: 20 }} </p>
   <div class="meta">
-    {{event.eventdate | date_to_string: "ordinal", "US"  }} 
+    <span class="locale-date">{{ event.eventdate | date_to_rfc822 }}</span>
     &middot;
     {% if event.online %} Online {% endif %}
     {% if event.location.city %} {{event.location.city}} {% endif %}
--- a/assets/js/mod/locale-date.js
+++ b/assets/js/mod/locale-date.js
@@ -1,0 +1,37 @@
+define(['jquery'], function($) {
+    $('.locale-date').each(function() {
+        var date = new Date(this.innerText);
+        this.replaceWith(date.toDateString());
+    });
+
+    $('.locale-datetime').each(function() {
+        var p = this.parentElement;
+        var date = new Date(this.innerText);
+        var local_tz = Intl.DateTimeFormat().resolvedOptions().timeZone;
+        var event_tz = this.dataset['tz'];
+
+        $('<h3>In your time zone (' + local_tz + ')</h3>').prependTo(p);
+
+        this.replaceWith(date.toLocaleDateString());
+        $('<br />').appendTo(p);
+        $('<strong>Time:<strong>').appendTo(p);
+        p.append(' ');
+        p.append(date.toLocaleTimeString());
+
+        /*
+         * On-site events have a time zone. If the user time zone is different
+         * than the event one, also display the date and time of the event from
+         * the event timme zone.
+         */
+        if (event_tz && event_tz != local_tz) {
+            $('<h3>In the event time zone (' + event_tz + ')</h3>').appendTo(p);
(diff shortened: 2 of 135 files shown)


smortex:
FYI, there are still some events with an invalid time zone (same issue as #1616). For now I kept them unchanged because this PR change the way dates or formatted and we do not want to have to deal with merge conflicts when syncing branches. We can fix them when this is merged.

events/index.html
```diff
@@ -6,8 +6,8 @@
 notice: true
 ---
 
-{% assign now =  site.time |  date_to_xmlschema %}
-{% assign future_events = site.events | where_exp: "event","event.eventdate  > now" %}
+{% assign now = site.time %}
+{% assign future_events = site.events | where_exp: "event","event.eventdate > now" %}
```
smortex, line 10: By using a parsable date for `event.eventdate`, we can use native comparison rather than text comparison to filter / sort dates.

_includes/event.html
```diff
@@ -2,7 +2,7 @@
   <a href="{{ event.url }}">{{event.title}}</a>
   <p> {{ event.content | strip_html | truncatewords: 20 }} </p>
   <div class="meta">
-    {{event.eventdate | date_to_string: "ordinal", "US"  }} 
+    <span class="locale-date">{{ event.eventdate | date_to_rfc822 }}</span>
```
smortex, line 5: We format the date as defined by RFC822 so that Ruby native date is serialized in a predictable way JavaScript can parse.

assets/js/mod/locale-date.js
```diff
@@ -0,0 +1,37 @@
...
+        /*
+         * On-site events have a time zone. If the user time zone is different
+         * than the event one, also display the date and time of the event from
+         * the event timme zone.
+         */
+        if (event_tz && event_tz != local_tz) {
+            $('<h3>In the event time zone (' + event_tz + ')</h3>').appendTo(p);
+            $('<strong>Date:<strong> ').appendTo(p);
+            p.append(' ');
+            p.append(date.toLocaleDateString(undefined, {timeZone: event_tz}));
```
smortex, line 30: The first parameter allows to change the language.  By not setting it, the user locale representation will be used.

While we can hardcode it to be `en-US` and match the format used in the US on the website which is written in English for consistency, I chose not to do so because "6/1/23" hardly makes sense for me and the default for my locale, "01/06/2023", is immediately readable and understood by my brain.  My guess is that it is the same for people with other time formats, so I preferred legibility over consistency.

krisfreedain:
This is a fantastic idea @smortex - thank you! 
@nateynateynate - can you run through this? Looks good to me but I'd appreciate you giving it a spin as well.

nateynateynate (approved):
Ok, I'm cool with this. This is rad, @smortex!

nateynateynate:
Going to stage this first - waiting to collaborate with another maintainer.

nateynateynate:
Hey @smortex - looks like when events are made with time formats that don't conform, the build process fails when comparing a string to a time. Any chance you can add a quick check for types and/or fall back on the default value?

smortex:
> Hey @smortex - looks like when events are made with time formats that don't conform, the build process fails when comparing a string to a time. Any chance you can add a quick check for types and/or fall back on the default value?

Yeah, Jekyll cannot compare non-comparable objects (things that are formatted like date and times or converted to a date object, but the rest stays as String).  The backtrace sucks however, and does not point to the malformed line.

Maybe we can add a check in the pipeline:

```
grep -Er --files-without-match '^eventdate: [0-9]{4}-[0-9]{2}-[0-9]{2} [0-9]{2}:[0-9]{2}:[0-9]{2} [-+][0-9]{4}$' _events
```

I was about to open a PR for this but saw the code was revert.  I do not know how I should handle this: revert the revert, add the check and fix the code in a new PR?

Issue #1169, opened 2022-11-16
[Website update] Social sharing previews - blogs, individual pages, community members


krisfreedain:
### Is your feature request related to a problem? Please describe.

Every time a page on our site is shared on twitter, the preview appears with the same description and image (the OpenSearch logo). We should be able to specify more of the presentational details on an as-needed basis.

There should also be default images for blog posts that don't have a featured image based on the category front matter.

### Describe the solution you'd like

Implement some jekyll logic in the header includes such that it will check the page for any defined twittercard frontmatter. If any is defined, honor it. Otherwise, use the default.

If a post doesn't have a featured image, use a default based on the category.
The content of the twittercard should be the first paragraph of content on the blog, or perhaps use the excerpt frontmatter if it exists.

Let's provide a few arrangements of defaults that we can pick between as-needed.

### Additional context

https://developer.twitter.com/en/docs/twitter-for-websites/cards/overview/abouts-cards

krisfreedain:
11/4 Hannah has a mock up cards for general use. - will need to build this in a comment here on GH. 
Hannah finished GitHub training. Will begin the "self serve" process of creating a account. - @nateynateynate  will pick thru this and begin coding this in the blog template.

2022-1025 - this may require a 'featured image' for every blog post. Need to research that Jekyll can handle / have option for this. Hannah will work with Daryll/Nate/Kris on this one.

10/2022 Site default card to large image and we will make that unique.
Integrated into publishing process so we can make sure we provie the image URL to the twitter card.- Manual to start. More frontmatter into the top of the blog. Blog owners choose the image they want for the blog. 

action item:

-A Github issue to fix twitter cards.
-create default twitter card.

krisfreedain:
@nateynateynate - this is WIP - will connect w/ him next week on status

krisfreedain:
@nateynateynate will be working on this next
@Uselmh will create a default 800 x 418 (image size ratio) - for shares/blogs that do not have their images

astephanus:
@krisfreedain @nateynateynate According to the Twitter documentation there is special behavior regarding the description field that we need to consider. It is not displayed on iOS and Android mobile apps, and for the Web it is truncated to only 3 lines for both timeline and for the expanded Tweet. FYI: https://developer.twitter.com/en/docs/twitter-for-websites/cards/overview/summary-card-with-large-image

kaimmej:
### Current
[...]

PR #1642, opened 2023-06-01
Blog twitter cards 1169

astephanus:
### Description
Implements Twitter Card meta data by both correcting, and completing the previously incorrect, and incomplete implementation. 

- Correctly accounts for field length restrictions. 
- Allows for various ways to define `twitter:description`; either explictly using `twittercard.description` Front Matter with a fallback to `excerpt`, and then to `meta_description`. 
- Allows for specification of `twitter:creator` with a fallback to the first author (The fallback should to be discussed.)
- Allows for explicit specification of `twitter:site` using a Front Matter property named `twittercard.site` with a default of `@OpenSearchProj`.
- Template logic sets the Twitter Card type to `summary_large_image` for content with an image and `summary` for content without an image.
- Template logic will use `twittercard.image` if defined, or `feature_image` if not defined. 
- Adds `twittter:image:alt` for accessibility support by using `twittercard.image_alt` in the Front Matterr.
- Removes a default image from the `head.html` template in accordance with Twitter's best practices about not using the same image for different content.
- Adds values for `twittercard.description` taken from first paragraph page content for blog posts for posts that did not already have a value defined for `excerpt`. Markdown code has been stripped from those values, and some have been  subjectively truncated based on the fact that Twitter allows only for 200 characters across 3 lines (Only for content that was obviously very, very much longer). Values are still explicitly truncated in the `head.html` template as not to violate Twitter's rules.

Solution requires validation through Twitter's Card Validator (https://cards-dev.twitter.com/validator). It is not of this moment certain how to do so from localhost. 

Values that have been added for `twittercard.description` should be reviewed by editorial staff.
 
### Issues Resolved
https://github.com/opensearch-project/project-website/issues/1169

### Check List


nateynateynate:
I'll check this out best I can, but we might need to stage this and then run some twitter card previewers before we hit production just for a few sets of eyes. 

@krisfreedain : Mind working with me after I put eyeballs on this?

_includes/head.html
```diff
@@ -20,31 +20,42 @@
...
-      {% assign author_short_name = page.authors | first  %}
-      {% assign author = site.authors | where: 'short_name', author_short_name | first %}
-      {% if author.twitter %}
-        <meta name="twitter:site" content="@{{ author.twitter }}" />
-      {% endif %} 
+    {% if page.twittercard.image or page.feature_image %}
+    <meta name="twitter:card" content="summary_large_image">
+    {% else %}
+    <meta name="twitter:card" content="summary">
     {% endif %}
```
nateynateynate, line 28: I believe that we can also specify `player` for the meta keyword `twitter:card` - we may want to some day tweet something that plays automagically for people on twitter.

astephanus: @nateynateynate Would a reasonable solution be to accept a specific card type indicated in the Front Matter and then fall back accordingly if not specified?

astephanus: @nateynateynate I have updated the `head.html` template with the player type logic.

nateynateynate (approved):
LEts rawk.

PR #1669, opened 2023-06-08
Revert "Localize the date and time of events"

nateynateynate:
Reverts opensearch-project/project-website#1640

@AMoo-Miki - We've got some kind of divergence between when we perform a `bundle exec jekyll serve` to test things locally and what the build process performs. This worked just fine locally but failed during CI/CD. 

Can you have a good close look at this one?


krisfreedain (approved):
Reverting while we figure out why it won't build on the backend

PR #1674, opened 2023-06-09
Detect malformed event dates

smortex:
Detect malformed date for events so that we can fix them quickly.


smortex:
Ping @nateynateynate :wink:

nateynateynate:
@smortex You're a rockstar! Thanks for this.

smortex:
I merged the master branch in this PR in order to fix the events added recently with #1695.  Their time-zone info seems to be wrong, but I will handle this in another PR when I am done with this one and comparing the date on meetup and the website is more straightforward :wink:.

smortex:
:wave: I am back from some time off!

I added another merge commit to adjust the recently added events.

smortex:
I just pushed one more commit to fix the TZ errors in the future events (comparing the TZ that is used in the event description and the one used on meetup).

-----

While doing this, I saw some inconsistencies on meetup for these events you host @krisfreedain :

* https://www.meetup.com/opensearch/events/294451568
* https://www.meetup.com/opensearch/events/294451571
* https://www.meetup.com/opensearch/events/294451577

When showing these events, meetup hardcode a UTC-7 time-zone, while the other events use the browser time-zone. I guess the config is slightly different (attached to a location maybe?), and as the event is online, the user time zone should be preferred.

krisfreedain:
thanks @smortex - yeah, they seem to hardcode. I'll look through the Meetup settings and see if there is a way to enable browser time-zone instead though.

_events/2023-0718-2-9-release-meetings.markdown
```diff
@@ -1,12 +1,11 @@
 ---
 
-eventdate: 2023-07-19T15:00
-tz: UTC -8
+eventdate: 2023-07-18 15:00:00 -0700
```
smortex, line 3: According to the schedule and the dates bellow, it is 2023-07-18.

_events/2023-0718-2-9-release-meetings.markdown
```diff
@@ -1,12 +1,11 @@
...
 title: OpenSearch 2.9.0 Release Meetings - 2023-07-18
-#online: true
-#signup:
-#    url: https://www.meetup.com/opensearch/events/tvcvftyfclbcc/
-#    title: Join on Meetup
+online: true
+signup:
+  url: https://us02web.zoom.us/j/83030994833
+  title: Join on Meetup
```
smortex, line 8: This was copied from the links at the bottom of the description to avoid a useless empty button on the side.

dtaivpp (approved):
Lets see if this breaks prod

PR #1798, opened 2023-07-19
Pre commit hook fix

dtaivpp:
### Description
Fixing the shell script that fails the events pre-commit hook. 
Adds local linting using #1674

### Check List

### DIFF
--- a/_scripts/_malformeddates.sh
+++ b/_scripts/_malformeddates.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-malformed_event_dates=$(grep -Er --files-without-match '^eventdate: [0-9]{4}-[0-9]{2}-[0-9]{2} [0-9]{2}:[0-9]{2}:[0-9]{2} [-+][0-9]{4}$' _events)\
+malformed_event_dates=$(grep -Er --files-without-match '^eventdate: [0-9]{4}-[0-9]{2}-[0-9]{2} [0-9]{2}:[0-9]{2}:[0-9]{2} [-+][0-9]{4}$' _events)
 if [ -n "${malformed_event_dates}" ]; then
 echo "Malformed event date in:" >&2
 echo "${malformed_event_dates}" >&2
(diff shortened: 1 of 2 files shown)


_scripts/_malformeddates.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
```
smortex, line 1: This script looks portable bourne shell, so we can rely on `/bin/sh` ;-)

```suggestion
#!/bin/sh
```

If we really really want bash, I suggest the following to cope with systems where bash is installed in another directory:

```suggestion
#!/usr/bin/env bash
```

smortex:
Looks like we can add more checks:

* When `online: true`, we should not have a `tz:` entry.
* When we have a `tz:` entry, it sould use the format `Continent/City` which avoid mistakes because of daylight saving time.

_scripts/_malformeddates.sh
```diff
@@ -0,0 +1,8 @@
...
+
```
smortex, line 2:
```suggestion
online_and_timezone=$(grep -Er --files-match '^online: true$' | xargs grep --files-with-match '^tz:')
if [ -n "${malformed_event_dates}" ]; then
  echo "Online events should not have a time-zone.  Offending files:" >&2
  echo "${malformed_event_dates}" >&2
  exit 1
fi
malformed_timezones=$(grep -Er --files-match '^tz:[^/]*$')
if [ -n "${malformed_timezones}" ]; then
  echo "Time-zones should be in the form 'Continent/City'.  Offending files:" >&2
  echo "${malformed_timezones}" >&2
  exit 1
fi
```

smortex: Thinking more about this, and after discussion with @nateynateynate I wonder if we should not always accept a `tz: Foo/Bar` (aka Time zone identifier) even for online events, and when one is found verify that the event time in `eventdate` use the same time-zone offset that that date when using the provided Time zone identifier.  This would allow to make the build fail when we encounter inconsistencies.

#### A valid example
```
eventdate: 2023-02-08 08:00:00 -0800
nz: America/Los_Angeles
```

#### An invalid example

Same time of the day, but 6 months later:

```
eventdate: 2023-08-08 08:00:00 -0800
tz: America/Los_Angeles
```

The above is **not** valid: on that date, the TZ offset in America/Los_Angeles is UTC-7 (because of Daylight Saving), and the above date in that time zone is therefore `2023-08-08 09:00:00 -0700`.  If "08:00 am" was intended, it should have been written with the correct time-zone offset:

```
eventdate: 2023-08-08 08:00:00 -0700
tz: America/Los_Angeles
```

dtaivpp: Seems like a reasonable suggestion.

PR #1802, opened 2023-07-19
Remove time-zone information for online events

smortex:
When an event is online, we show the start date and time in the user time-zone.

When a `tz` is present, the date and time of the event in that local time zone is also shown to the user, but this is only intended to on-side events: it is intended to help the user better plan their trip.

| It currently show confusing information (local build): | With this change, less confusion for an online-event: |
|----|----|
| current | after |

### DIFF
--- a/_events/2023-0718-2-9-release-meetings.markdown
+++ b/_events/2023-0718-2-9-release-meetings.markdown
@@ -2,7 +2,6 @@
 
 
 eventdate: 2023-07-19 15:00:00 -0800
-tz: UTC -8
 title: OpenSearch 2.9.0 Release Meetings - 2023-07-18
 online: true
 signup:
--- a/_events/2023-0724-dev-triage-security.markdown
+++ b/_events/2023-0724-dev-triage-security.markdown
@@ -1,7 +1,6 @@
 ---
 
 eventdate: 2023-07-24 12:00:00 -0800
-tz: UTC -8
 title: Development Backlog & Triage Meeting - Security - 2023-07-24
 online: true
 signup:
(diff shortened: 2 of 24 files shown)


smortex:
Cc @dtaivpp @nateynateynate

_events/2023-0718-2-9-release-meetings.markdown
```diff
@@ -2,7 +2,6 @@
 
 
 eventdate: 2023-07-19 15:00:00 -0800
```
smortex, line 4: Also, the page says "July 19th - 3PM PT" bellow, and PT is currently PDT, that is -0700 and not -0800…

```suggestion
eventdate: 2023-07-19 15:00:00 -0700
```

TZ abbreviations are evil :smiling_imp: !

dtaivpp:
@smortex thanks! That explains that. Also, these don't render on firefox it seems like. Web not moved to prod yet because of that. If you don't have time to look at it I'll be digging in tonight.

smortex:
> Also, these don't render on firefox it seems like.

What? I am using Firefox and it works on my computer :confused:.  For completeness, I run Firefox 102.13.0esr (64 bits) on Debian 11.  The process has a TZ environment variable:

```sh-session
romain@desktop-fln40kq ~ % pgrep firefox
2770
romain@desktop-fln40kq ~ % cat /proc/2770/environ | tr '\0' '\n'  | grep TZ
TZ=Pacific/Tahiti
```

I'll try to fiddle with my conf to see if I can produce unexpected results. Can you provide a screenshot of the disaster and info about the version you are running / TZ on your system?

dtaivpp:
@smortex okay so I'm not sure if I need to be mad at firefox or our cloudfront but it seems like it was just a caching issue -_-

nateynateynate (approved):
LGTM. Thanks so much @smortex!

PR #1805, opened 2023-07-20
Fix the TZ offset of future events

smortex:
PT is currently PDT (UTC-7), and not PST (UTC-8).  It looks like the wrong offset was used, inducing a off-by-one error in the displayed times.


nateynateynate (approved):
Rockstar. Thanks @smortex

PR #1808, opened 2023-07-20
2.9 Release date updates

krisfreedain:
### Description
Updated the release information
 
### Issues Resolved
n/a

### Check List
- [x] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the BSD-3-Clause License.

PR #1812, opened 2023-07-21
Fix event link in the sidebar

smortex:
This seems to have been changed by mistake in a464fa34b828bf44cab69ca3912ccbcb80312fdd.

While here, use the PDT time-zone abbreviation rather that PT which is PDT or PST depending on the date.


smortex:
Cc @krisfreedain @nateynateynate as this is a follow-up to #1808.

nateynateynate (approved):
Thanks for the quick fix!

PR #1992, opened 2023-09-12
As much as I would like to have a workshop that goes from 9am to 11pm…

nateynateynate:
…, I'm afraid this just isn't in the cards.

Let's fix that to say 11am.

### Description
Typo in workshop session says 11pm as the end time. 
 

### Check List


krisfreedain (approved):
LGTM

PR #1999, opened 2023-09-14
Platform Pages and Redesigned Nav / Footer

astephanus:
### Description
Adds new Platform Landing Page.
Adds new Platform Security Analytics page with Platform Category template.
Adds new design for site navigation header.
Adds new design for site footer.

### Check List


AMoo-Miki:
Do to the huge number of files changed, let's make sure we have a few of FEE folks review this before merging.

dtaivpp:
1. For the `Community Voices` page on mobile there are 3 nested elements that scroll independently which can be challenging to activate on mobile. 

2. For the OpenSearchCon hero banners we need to have some way to ensure we don't upload wide images. Maybe we need a mechanism for swapping these images depending on the layout. 
Screenshot 2023-09-25 at 4 15 18 PM

3. Download on OpenSearch and Checkout on GitHub overflow past the website boundaries on mobile. 
Screenshot 2023-09-25 at 4 18 37 PM

4. Nav bars flow behind main content window on mobile: 
Screenshot 2023-09-25 at 4 19 36 PM

5. Are we supposed to be able to search on mobile view? 

6. Testimonials page seems to have broken images on the testimonials. 
Screenshot 2023-09-25 at 4 21 59 PM

7. Some page titles overflow in unexpected ways. 
Screenshot 2023-09-25 at 4 23 12 PM

_sass/_redesign-platform.scss
```diff
@@ -156,138 +233,748 @@ $platform-page-content-reduce-padding-width: 1199px;
...
+                    left: 0;
+                    right: 0;
+                    bottom: 0;
+                    margin: 0;
+                    padding: 0;
+                    width: 100%;
+                    height: 100%;
+                    &.landing-page-hero--images--image__desktop {
+                        > img {
+                            object-fit: fill;
```
AMoo-Miki, line 654: There are 10 uses of `object-fit: fill` in 
* `_sass/_redesign-platform.scss`
* `_sass/_redesign-platform.scss`
* `_sass/_redesign-platform.scss`

which will result in distorted and unpleasant rendering. `object-fit: cover` will be better.

That said, I am not a fan of the `position: absolute` and the fixed height and width (dictated by a fixed aspect ratio).

_includes/icons.html
AMoo-Miki, line 1: * `arrow-left` and `arrow-right` are using different colored masks and different colors for the arrows. As a result, the arrows look different.
* The colors are not contrasting enough and are often lost on the background; I would recommend changing the fills for the masks and the fills for the paths to `#FFFFFF`.

_includes/header.html
```diff
@@ -17,73 +17,152 @@
...
+      invisibleLink.classList.remove(invisibleClassName);
+      invisibleLink.classList.add(visibleClassName);
+      
+      const childUL = getSubMenu(toggle);
+      childUL?.toggleAttribute?.('expanded');
+    }
+    const topNavigationToggleButtons = document.querySelectorAll('#top .opensearch-toggle-button--wrapper');
+    for (let i = 0; i < topNavigationToggleButtons.length; ++i) {
+      const button = topNavigationToggleButtons[i];
+      initializeCustomMenuHeights(button, i);
```
AMoo-Miki, line 164: This should be handled by CSS using media queries.

_includes/header.html
```diff
@@ -17,73 +17,152 @@
...
+      button.addEventListener('click', onToggleButtonClick);
```
AMoo-Miki, line 165: One delegate should handle all of these; adding too many event listeners could slow the page on certain platforms.

_includes/copy_banner.html
```diff
@@ -4,22 +4,6 @@
         {% assign primary_title = page.primary_title | default: primary_title | default: '' %}
         {% assign primary_link_url = layout.primary_link_url | default: page.url %}
         {% if primary_title.size > 0 %}<h1><a href="{{ primary_link_url }}">{{primary_title}}</a></h1>{% endif %}
-
-        <div class="search">
```
AMoo-Miki, line 8: `.search` had `z-index: 2` forcing the search results appear on top of `.search-overlay`; that is lost with the new design.

_includes/header.html
```diff
@@ -17,73 +17,152 @@
...
-        </svg>            
+      {% include icons.html type='opensearch-logo-default-1' %}
```
AMoo-Miki, line 20: Can we just move this to an external svg and load it as an image?

_includes/icons.html
astephanus: This is resolved

_sass/_redesign-platform.scss
astephanus: I agree. We need a specification for how images behave responsively. Whether to maintain aspect ratio with rules about how to position, and what to fill in the space, or to crop. I have reached out for design guidance on this. The really high touch graphics look great at or near their source geometry, but across variable viewports things get dodgy.

_includes/header.html
astephanus: Agreed. The event listener is one static function reference, and there are only 4 buttons, and I have hardcoded the submenu heights for the custom CSS property to be applied as `--expanded-height`. I did this quick and dirty to get it done, however I have this resolution prioritized as a fast follow; along with a runtime determination of the height. This is for being able to do a smooth height transition.

_includes/header.html
astephanus: That's interesting. I'm not sure what the most optimal approach would be. The idea here is that we have a Figma pattern library with named icons, and the `icons.html` include file can be used to pull in those icons where they're needed. Would the SVG data being external improve performance, either at build time or at runtime?

AMoo-Miki: Using the icons.html this way practically inlines the source of the SVG, making the pages heavier. If they were external assets, they would be cached for subsequent uses.

nateynateynate (approved):
LEEEEEEEROYYYYYY JEEEEENKEEEEEINS

PR #2033, opened 2023-09-21
[OSCI-23] Add `vchrombie` as an author

vchrombie:
### Description
Add `vchrombie` as an author
 
### Check List


dtaivpp (approved):
If you noticed we have a few errors that were checked by our `style workflow` step. This is okay in your case as they were all false positives. Looks good!

vchrombie:
> If you noticed we have a few errors that were checked by our `style workflow` step. This is okay in your case as they were all false positives. Looks good!

Thanks for the review.

dtaivpp (approved):
Looks good!

PR #2288, opened 2023-09-27
Fix OpenSearchCon 2023 info

smortex:
The event does not include the location information (city, country, TZ) so the event page lacks a lot of information.

While here, adjust the event logic so that events which started and are not finished at the time of rendering appear both in the future events (because it is what is displayed on the event page and the event is still in progress) and the past events.

Also adjust the event template with actual sample data in comments to make it more obvious for people using it what they need to adjust.

| Before (Currently online) | After |
|--------|-------|
| image | image |
| image | image |

### DIFF
--- a/_events/2023-0927-opensearchcon.markdown
+++ b/_events/2023-0927-opensearchcon.markdown
@@ -1,8 +1,13 @@
 ---
 
 eventdate: 2023-09-27 08:00:00 -0700
+enddate: 2023-09-29 20:00:00 -0700
 title: OpenSearchCon - 2023
-online: 
+online: false
+tz: America/Los_Angeles
+location:
+    city: Seattle
+    country: USA
 signup:
     url: https://opensearchcon2023.splashthat.com/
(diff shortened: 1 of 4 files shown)


nateynateynate:
Thanks for this @smortex - sorry we didn't get to it in time. I think we'll end up with a different approach for our digital assets next year, but if we end up making an event for it I'll make sure we add the appropriate timing information. :-)

smortex:
No worries, I try to check from time to time if there are inconsistencies in order to fix them like this one: even if there are "sample" events to use as a starter, I bet most people will pick an existing event that looks-like the one they want to add and copy it as a base, so fixing the old events metadata will help avoid mistakes in the future.

smortex:
Just fixed this PR while working on #2418: Seattle is in the TZ America/Los_Angeles not America/Chicago as was previously set (and visible on screenshots).  I did not re-do the screenshots, but the commits are amended to have the right information.

Issue #2308, opened 2023-09-29
[BUG] Featured Blog Post Excerpt is Malformed


astephanus:
**Describe the bug**
The homepage featured blog post excerpt will not be properly styled depending on how it is authored.

**To Reproduce**
1. Go to https://opensearch.org/
2. Scroll down to the Featured Blog posts hero.
3. The featured blog post preview with excerpt will use more vertical space that it is designed (at least with the current blog post of https://opensearch.org/blog/remote-backed-storage/). 

**Expected behavior**
The preview should fit within its bounding box, and the excerpt text should be line-clamped at 5 maximum lines with an ellipsis if necessary.

ReinGrad:
* Adding overflow: hidden and text-overflow: ellipsis to the excerpt container to enforce line clamping
* Adjusting line heights or margins to condense vertical space
* Modifying excerpt generation logic to truncate at a certain character count

astephanus:
@ReinGrad This was already solved a week ago in PR https://github.com/opensearch-project/project-website/pull/2309. I've left the issue open as the PR has yet to be reviewed and merged.

astephanus:
Resolved in PR https://github.com/opensearch-project/project-website/pull/2309.

Issue #2295, opened 2023-09-28
[BUG] New Homepage Featured Blog Post Does Not Always Show Author Name


astephanus:
**Describe the bug**
The newly released homepage "Featured Blog Posts" preview element displays author name + Twitter handle. Depending on what content is featured the author name does not always appear leaving an empty space before a "|" followed by the `@OpenSearchProj` Twitter handle.

**To Reproduce**
1. Go to https://opensearch.org/ on a desktop browser at full width (the component collapses into something else responsively for smaller device screens).
2. Scroll down to Featured Blog Posts
4. Observe that the featured blog post may, or may not show an author name. At this moment the "Exploring OpenSearch 2.10" post is featured written by James McIntyre, but James's name is not appearing.

**Expected behavior**
Author name should always appear.

astephanus:
Resolved in https://github.com/opensearch-project/project-website/pull/2309

PR #2309, opened 2023-09-29
Fix homepage featured blog excerpt format

astephanus:
### Description
Solves for the condition when a featured blog post excerpt on the homepage contains an explicit `excerpt` value in its front matter not being properly formatted as is when Jekyll automatically uses the post's first paragraph.

Also implements a design change that removes the homepage featured blog post author name and twitter handle and instead puts in its place the post category icon and publish date. This obsoletes Issue #2295 

Also adds dynamic sub-nav menu height determination for smooth expand / collapse transitions instead of the pre-existing hardcoded values. This will prevent the need to update the hardcoded values every time navigation items are added or removed from the `_data/top_nav.yml`, or should styles for the menu items change.
 
Also adds a fix for seeing extra height scrollable height in the new carousel component on mobile.
### Issues Resolved
https://github.com/opensearch-project/project-website/issues/2308
https://github.com/opensearch-project/project-website/issues/2295

### Check List


krisfreedain (approved):
thanks @astephanus

PR #2311, opened 2023-10-01
Fixed issue 2302 (pls ignore previous PR)

### LINKED ISSUE #2302: [BUG] New Footer Is Missing Left / RIght Spacing
**Describe the bug**
The new footer design does not have the correct left / right page margins as other page elements.

**To Reproduce**
1. Go to https://opensearch.org/
2.  Scroll down to the bottom footer
3. Notice the 3 column footer (fullwidth desktop) appears to be misaligned horizontally. Also compare to the design in Figma, and notice that it does not match. 
4. The same issue is observable for 2 and single column responsive breakpoints.

ANANTH-SWAMY:
### Description
It fixes the alignment issues of the footer highlighted in the issue #2302 
 
### Issues Resolved
#2302 
### Check List

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -6,13 +6,15 @@
       <div class="subfooter--flex-wrapper">
         <div class="subfooter--columns-wrapper">
           {% for column in site.data.footer.columns %}
-            <div class="subfooter--columns-wrapper--column">
-              <h4>{{ column.title }}</h4>
-              <ul>
-              {% for link in column.links %}
-                <li><a href="{{ link.url }}">{{ link.title}}</a></li>
-              {% endfor %}
-              </ul>
+            <div class="subfooter--columns-wrapper--container">
+              <div class="subfooter--columns-wrapper--column">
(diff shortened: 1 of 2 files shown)


astephanus (changes requested):
@ANANTH-SWAMY do you have access to the Figma design for reference? The changes here eliminate the responsive flex-box rules for `.subfooter--columns-wrapper`'s flex items, and break the responsive design, and maintains that the footer will always be three columns, and not collapse into two on somewhat more narrow screens (such as tablets), and finally into one column on phones. The problem sited by issue 2302 is that the left-right padding within the overall page `.container` are improperly applied across all breakpoints. By adding the `*--container` block element, the `flex-*` rules in the nested `*--column` block element are being ignored. Look at how the left and right `padding` rules are being applied to other components above the footer for reference, for example on the new homepage `.solutions-card-grid`, and the SCSS Mixin named `page-element-padding` in `_redesign-layout.scss` for guidance.

PR #2415, opened 2023-11-06
Fixes header navigation menu mobile bug; removes dependency on rem units

astephanus:
### Description
Fixes header navigation menu mobile bug.
Removes dependency on rem units.
Fixes a bug in the responsiveness of the homepage hero banner carousel that I had caused and only partially fixed recently. Now it's fixed all the way.

### Check List


nateynateynate (approved):
Let's get this out there. It impacts our mobile view, and during reinvent week we want our mobile representation to be polished.

PR #2418, opened 2023-11-08
Fix more events metadata

smortex:
This is a follow-up to #2288 (merging this PR will automatically marked that older PR as merded).  More events have incomplete / invalid information.  Fix it.


krisfreedain (approved):
thanks @smortex

PR #2431, opened 2023-11-16
Fix OpenSearch Dashboards Developer Office Hours

smortex:
The event was probably using the wrong time-zone information (probably
due to daylight saving that recently changed).  While here, ensure the
join button was functional and fix its label.

The event is passed, but it will hopefuly help if it is taken as a
template to create a new event in the future.

Signed-off-by: Romain Tartière <romain@blogreen.org>


krisfreedain (approved):
thank you @smortex!!

PR #2451, opened 2023-11-27
Cancel Nov 28 community meeting

krisfreedain:
### Description
Canceling Nov 28 Community Meeting - much of the community's attention is on re:Invent this week.
 
### Issues Resolved
n/a

### Check List


nateynateynate (approved):
Community meeting cancelled.

smortex:
:thinking: I wonder if instead of removing such events we should add `[CANCELED]` to the title, and ~strike-out the content~.

My experience today: I remembered that there was an event tomorrow and wanted to check the time, but could not find it.  I was confused.  I though about looking in the GitHub history and found the event was removed by this PR.  Adjusting the event as described above would have helped me to understand quickly what was going on.

Sure, no need to re-add it, but maybe if it happen again in the future that can "help" someone :-)

krisfreedain:
@smortex - I definitely see your point @smortex and agree - we'll work to change that in future cancellations - thank you

PR #2422, opened 2023-11-13
Fix date of next community meeting

smortex:
In my time zone, the website says "Date: 14/11/2023 Time: 12:00:00", on
meetup, I see "Tuesday, November 14, 2023 at 1:00 PM to Tuesday,
November 14, 2023 at 2:00 PM UTC−10".  Looks-like the TZ offset is not
the good one.

PR #2456, opened 2023-11-28
Improve detection of Time-Zone inconsistencies

smortex:
We have a track record of Time-Zone inconsistencies in the events of the
project website (e.g. #1616, #1805, #2288, #2418, #2422, #2431, #2449).

In order to detect these inconsistencies earlier, make the `tz`
mandatory and check that the Time-Zone offset match the Time-Zone name.
These checks require a working bundle so run it on GitHub actions and
keep the old pre-commit check to only check date formatting using basic
UNIX tooling.

Adjust the events templates to match what we expect.

While here, also check that non-online events have the needed
`location.city` and `location.country`.

Signed-off-by: Romain Tartière <romain@blogreen.org>


smortex:
This will hopefully avoid more work for @krisfreedain and @nateynateynate to review and merge my PRs to fix Time-Zone issues :sweat_smile:

Cc @joshuarrrr who also hit some Time-Zone related issue.

I tried to split each part in a separate commit to make it easier to follow the fix process.  As a bonus, we can see that the check script report less and less failures as we go.

Thanks!

krisfreedain:
WOW! Thank you @smortex!!
 
@astephanus - could you give this a look?

PR #2490, opened 2023-12-12
Removed Opster since they are part of another company now

jkowall:
Since they are part of a commercial entity that is not part of the OpenSearch ecosystem they have stepped back from the community per @segalziv 

### Description
Removes partnership page for Opster
 

### Check List


dtaivpp:
Feels wrong given their contributions to OpenSearch but we don't have a mechanism to move them to emeritus. I'll review this today thanks!

krisfreedain:
Before we do this, we would need to ensure @segalziv voice is captured in this conversation

jkowall:
We should keep the authors and people there, but the company no longer exists and is part of Elastic since the deal is closed. 

image

dtaivpp (approved):
Looks good

PR #2517, opened 2023-12-29
Configuring Anonymous Authentication in OpenSearch blog post

smortex:
I recently configured anonymous authentication in OpenSearch.  While the documentation has some information about this, there is no guide available that list all requirements, and I had to find information in multiple places and had to look at the code to fully understand what was needed.  Not great for a tasks that I assumed would be simple.

More recently, people asked on Slack how to enable anonymous auth, so maybe a blog post is welcome to document the process?


_posts/xxxx-xx-xx-configure-anonymous-authentication.md
```diff
@@ -0,0 +1,101 @@
+---
+layout: post
+title: Configuring Anonymous Authentication in OpenSearch
+authors:
+  - smortex
+date: 2023-12-28 16:20:00 -1000
```
smortex, line 6: We'll need to adjust this and rename the file accordingly.

PR #2542, opened 2024-01-26
WIP - DO NOT MERGE - Meetup api integration

astephanus:
### Description

- Adds a NodeJS script for retrieving OpenSearch Project events for each group associated with the project from the Meetup GraphQL API. 
- Adds a NodeJS script for dynamically creating the Calendars Collection entries based on the state of the Events Collection. This replaces the former mechanism for doing this with a combination of using the `assembele_calendar_pages.sh` shell script to copy the contents of `_calendars/_sample.md` from a start year and month to an ending year and month, and then editing the 3 data points manually that couldn't be determined using a shell script; which was error prone, and none-specific as to why calendar collection items exist or do not.
- Adds additional node_modules dependencies for managing environment variables ("dotenv"), and using JSON Web Tokens ("jsonwebtoken").
- Adds two named script entries to the "scripts" block of `package.json` for running the Meetup events aggregation script, and for running the calendar collection management script.
- Adds missing event date-time localization to the individual event page template.
- Corrects logic error in the calendar layout template for dealing with leap years.
- Adds mock data for Meetup API query response bodies that was copied from the Meetup API Documentation GraphQL Playground.
- Removes an unused event category name from the sample events collection markdown file.
- Adds `node_modules` to `.gitignore`.

Because there are 2 new Node dependencies it is necessary to run `npm install` from the root directory to install the dependencies. Then in order to run the scripts to aggregate the events, and then create the correct calendar collection items run the following from the root directory:

1. `npm run meetup:fectch`
2. `npm run calendars:create`

Following successful execution the additional Events collection files will be visible in the `_events` directory, and the corresponding Calendar collection files will be visible in the `_calendars` directory if any changes are made (there should be calendar items extending to `2025-01.md`.

### Needs Help With

The current state of the Meetup API JWT flow is untested. The application logic is proven using mock data that was copied from real queries group events using the Meetup API documentation GraphQL Playground. Meetup API client IDs and keys need to be created and managed as environment variables. See the `.env` file. Also, once that information is available the functionality of the API client need to be tested by changing the two calls to the `mock__queryForGroupEvents` function to call the `queryForGroupEvents` function instead. The business logic processing the response should not need to be modified.

The JSON Web Token behavior needs guidance.

There seems to be a choice between managing known OpenSearch Project associated group names to query for all of their events, and moving the root of the GraphQL query schema up to the level of the OpenSearch Project Pro Network and either querying for all associated groups and then their events, or querying for all events directly. However, this has not been able to be proven since querying the relevant fields on the Pro Network schema returns an error using the Playground, and I suspect that the required admin credentials are required to prove these options as viable.

### Style Note

There are JavaScript class definitions in the Meetup script. They exist to define context and the expectations about what data is required, and not really for constructor functions, and type safety.

Liberal usage of JSDoc style doc-block comments, as well as verbose, descriptive, and hopefully meaningful variable and function names to attempt to explain how things work, and why choices were made.

### Check List


create-calendar-collection.js
```diff
@@ -0,0 +1,280 @@
...
+
+    // Since only the year and month, and not the day in the filename is of concern then
+    // it is reasonable to deduplicate the array.
+    const deDuplicatedFileNames = Object.values(filteredParsedFileNames.reduce((carry, current) => {
+        const joinedPair = current.join('-');
+        if (typeof carry[joinedPair] === 'undefined') {
+            carry[joinedPair] = current;
+        }
+        return carry;
+    }, {}));
```
AMoo-Miki, line 58: These 3 could be combined into one `reduce`: `3n -> n`

create-calendar-collection.js
```diff
@@ -0,0 +1,280 @@
...
+ */
+function parseYearMonthPairsFromFileNames(fileNames) {
+    const parsedFileNames = fileNames.map(fileName => {
+        const pattern = /^(\d{4})-(\d{2})/;
+        const matches = pattern.exec(fileName);
+        if (matches) {
+            const parsedYear = Number.parseInt(matches[1], 10);
+            const parsedMonth = matches[2].startsWith('0') ?
+                Number.parseInt(matches[2].charAt(1), 10) :
+                Number.parseInt(matches[2], 10);
```
AMoo-Miki, line 43: Checking if it starts with `0` isn't adding any value. Also, this forgives bad values like `20`.

It is almost guaranteed that this piece of code will not remain as is for the next 75 years so:
```
const pattern = /^(20\d{2})-(0[1-9]|1[0-2])/;
```
This will validate the correctness of `YYYY-MM` part. However, that still doesn't do what the comment says it will because it doesn't check the `DD`. We should either change the comment to relax the criteria or apply the strict criteria by not only parsing the numbers, but also validating the date. My sugestion would be a combination:
1. relax the criteria to `YYYY-MM##@` where `#` is just some number and `@` is a non-numeric character, and
2. change the pattern to `/^(20\d{2})-(0[1-9]|1[0-2])\d{2}([^\d]|$)/`.

That said, in the past we have had complications caused by poor choice of filenames. I would much rather read the content of the file and parse the actual date from there.

create-calendar-collection.js
```diff
@@ -0,0 +1,280 @@
...
+        const joinedPair = current.join('-');
+        if (typeof carry[joinedPair] === 'undefined') {
+            carry[joinedPair] = current;
+        }
+        return carry;
+    }, {}));
+
+    // Fill in the gap from HARD_START_DATE up to the [year, month]
+    // before the first entry. 
+    const [firstYear, firstMonth] = deDuplicatedFileNames[0];
```
AMoo-Miki, line 62: This assumes that `readdir` will sort the returned list.

create-calendar-collection.js
```diff
@@ -0,0 +1,280 @@
...
+    const deDuplicatedFileNames = Object.values(filteredParsedFileNames.reduce((carry, current) => {
+        const joinedPair = current.join('-');
+        if (typeof carry[joinedPair] === 'undefined') {
+            carry[joinedPair] = current;
+        }
+        return carry;
+    }, {}));
+
+    // Fill in the gap from HARD_START_DATE up to the [year, month]
+    // before the first entry.
```
AMoo-Miki, line 61: It would be great if we could add some more info here.

create-calendar-collection.js
```diff
@@ -0,0 +1,280 @@
...
+ * @returns {string}
+ */
+function createCalendarCollectionEntryData(yearMonth) {
+
+    const [year, month] = yearMonth;
+    const monthIndex = month - 1;
+    const previousMonthYear = monthIndex > 0 ? year : year - 1;
+    const previousMonthIndex = monthIndex > 0 ? monthIndex - 1 : 11;
+    const previousMonthEndDay = getDaysInMonth(previousMonthYear, previousMonthIndex);
+    const hoursOffsetBetweenLocalAndUTC = (new Date()).getTimezoneOffset() / 60;
```
AMoo-Miki, line 163: We should not rely on the local timezone for generating the content to make sure we don't get different content from machines running this script that don't have the same timezone.

create-calendar-collection.js
```diff
@@ -0,0 +1,280 @@
...
+    const formattedDateOfLastSaturday = `${lastSaturdayDate.getUTCFullYear()}-${lastSaturdayDateUTCMonth < 10 ? `0${lastSaturdayDateUTCMonth}` : lastSaturdayDateUTCMonth}-${lastSaturdayUTCDate < 10 ? `0${lastSaturdayUTCDate}` : lastSaturdayUTCDate}`;
+
+    const calendarEntryTpl = `---
+layout: calendar
+year: ${year}
+month_index: ${monthIndex}
+month_number: ${month}
+date_of_first_sunday: '${formattedDateOfFirstSunday}'
+date_of_last_saturday: '${formattedDateOfLastSaturday}'
+calendar_view_week_count: ${numberOfWeeksInCalendar}
```
AMoo-Miki, line 222: Correct me if i am wrong - I have not gone through all the code yet:

Are we doing the math in js for laying out the calendar and then letting jekyll build a page using that? If so, I would have preferred js to do everything and produce dumb html files for jekyll to enrich. This way, one language and one code builds the calander.

astephanus: This script automates the creation of Calendar Collection entries so that they don't have to be manually created, and are in sync with what needs to be shown in the calendar events pages. Technically speaking, they can be created manually. This is to prevent that. This doesn't actually build the calendar pages. Only the meta data that specifies what is shown. Not how it's shown. That is handled in `_layouts/calendar.html` and is easy for people who are not JavaScript programmers to  maintain in HTML, SCSS with Liquid logic.

PR #2583, opened 2024-02-14
Partner data line breaks. 

nateynateynate:
### Description
In preparation for implementing some kind of open authoring platform I need to get rid of some of the inconsistencies in our partner data. This just joines a few lines so that we don't have jekyll frontmatter that has linebreaks in it. 
 

### Check List


_partners/digitalpebble.md
dtaivpp, line 1: Is this a foobar? `DigitalPebble&apos;s`

dtaivpp (changes requested):

dtaivpp (approved):
Looks good!

_partners/digitalpebble.md
nateynateynate: Not necessarily. I think it's so that we can include apostrophes in single quoted front matter elements, but the doublequote makes more sense. Although I'm not sure if the same solution will still be required if we need to include a double-quote inside of a double-quoted string.

PR #2645, opened 2024-03-14
Merges Mobile UX fixes to production. 

nateynateynate:
### Description

A couple of updates. Mobile UX fixes, and some npm tasks to pre-populate new opensearchcon pages as well as archive old ones. 

### Check List


krisfreedain:
so now we're not going to show a "no image available" anymore?
@astephanus @nateynateynate 

Screenshot 2024-03-14 at 9 53 58 AM

nateynateynate:
> so now we're not going to show a "no image available" anymore? @astephanus @nateynateynate
> 
> Screenshot 2024-03-14 at 9 53 58 AM

Correct. Looks like there was a slack thread regarding this topic sometime around the time Maeve had to leave in which we decided it would be more efficient and look better not having to repeat that same image on the page so many times. I'm for the change - even if that image is cached on disk and it has to be read 100 times for every member that doesn't have an image, the user is still better off.

krisfreedain (approved):
thanks @nateynateynate @astephanus

PR #2982, opened 2024-06-19
Blog-feed updates - addresses small issues to the rendering of the blog cards

kaimmej:
Changes 

- Blog-categories are now concatenated with "," between multiple elements. 
- The date and blog-category are now on the same line.
- Small changes to the size of the featured post, it is now larger than the normal posts.
- Removed hover affect.


nateynateynate (approved):
Gets better every time!

PR #3000, opened 2024-06-21
fix - Social Sharing metatags - Pathing for Generic images was breaking

kaimmej:
I changed the logic in head-twitter-metatags so that it validates whether the category has an image and exists. If neither, it will set the category to default-category, and then pull the default image.

### Check List


nateynateynate (approved):
Ahh wonderful. I noticed the pieces of the head being duplicated. Great fix.

PR #3060, opened 2024-07-10
BUG -- new issue Project-assigning is not working. 

kaimmej:
we need to have the repo name and also the project name in order for the new issues to be assigned automatically to this project. In the github documentation, they show the numeric project numbers being used. The number "74" which is used in the feature-request template represents the "website update project".

The syntax are different between these two issues so that we can see if either will work.

Note: there are disclaimers saying that the person submitting the new-issue must have write-access to the repo for this auto-assign to the project to work. If neither of these work, likely we will have to scrap it and rely on github-workflows-autoassignissue if we want to implement this feature.

### Check List


nateynateynate (approved):
Awesome. Thanks Jon!

PR #3107, opened 2024-07-29
Adds new platform for download page, "Arch Linux" and the necessary i…

nateynateynate:
…nstructions to install via its native package manager.

### Description

#642 has been sitting around for a while. I was under the impression that there was some kind of clarification that was to be added to the download instructions regarding how the `opensearch` package is actually `opensearch-min`, but I think it's sat long enough. Here's the path I took: 

I added a dropdown to the OpenSearch and OpenSearch Dashboards sections of the download page to inlcude "Arch Linux" as a platform. Selecting any of the build variants will just show the command line instruction to install the `opensearch` package via `pacman` (`pacman -S opensearch`) 

Caveats: 
* The pacman package listing showed the only available architecture was x86_64, so I left out any other arch. 
* I didn't see pacman packages for the earlier release line of 1.3.x so I created no download artifacts for those. 

 
### Issues Resolved
#642 

### Check List

### DIFF
--- a/_artifacts/opensearch-2.15/x64/opensearch-2.15.0-archlinux-x64.markdown
+++ b/_artifacts/opensearch-2.15/x64/opensearch-2.15.0-archlinux-x64.markdown
@@ -1,0 +1,12 @@
+---
+role: daemon
+artifact_id: opensearch
+version: 2.15.0
+platform: archlinux
+architecture: x64
+slug: opensearch-2.15.0-archlinux-x64
+category: opensearch
+type: system-package
+archlinux_package_name: opensearch
+link: https://www.freshports.org/textproc/opensearch
+---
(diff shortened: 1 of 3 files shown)


hashworks:
> Selecting any of the build variants will just show the command line instruction to install the opensearch package via pacman (pacman -S opensearch)

As long as this doesn't confuse users since they need to install plugins via other packages.

> The pacman package listing showed the only available architecture was x86_64, so I left out any other arch.

Correct, we currently only provide x86_64. Others are planned (f.e. x86_64_v3, arm64), but this will take a while.

> I didn't see pacman packages for the earlier release line of 1.3.x so I created no download artifacts for those.

Also correct, Arch Linux will only ever provide the latest version of opensearch.

dtaivpp (approved):
Lgtm

PR #3416, opened 2024-10-29
Removed popper and switched the javascript include.

### LINKED ISSUE #3385: [BUG] Use self-hosted popper.js instead of the unpkg.com hosted JS.
### Describe the bug

As part of a recent CVE we're ensuring that no scripts are being loaded externally to the site from places like unpkg.com. 

Currently the only instance of unpkg I could find was for popper.js. Let's take care of that first. Then we can comb through for any other instances of externally hosted JS. 

### Expected behavior

All javascript should be hosted locally.

### Screenshots

image

"popperjs" being loaded from unpkg.com

### To Reproduce

Use developer console to view 'network' tab to see the domain we're pulling from.

### Additional Details

**Plugins**
Please list all plugins currently enabled.

**Screenshots**

**Host/Environment (please complete the following information):**

**Additional context**

kaimmej:
the bundle.min.js includes popper, which bootstrap uses to render certain elements through javascript. Specifically dropdown elements.

### Description
Removed the unpkg script that was loading the bootstrap javascript externally. 
Switched the include statement to the correct bootstrap bundle.min.js. 
 
### Issues Resolved
#3385 

### Check List

### DIFF
--- a/_includes/bootstrap-includes.html
+++ b/_includes/bootstrap-includes.html
@@ -1,3 +1,2 @@
 <link rel="stylesheet"  href="/assets/css/bootstrap.min.css">
-<script src="https://unpkg.com/@popperjs/core@2"></script>
-<script src="/assets/js/bootstrap.js"></script>
+<script src="/assets/js/bootstrap.bundle.min.js"></script>


nateynateynate (approved):
Looks good Jon! Thanks!

PR #3432, opened 2024-11-05
Remove logstash from downloads page and fix blog

derek-ho:
### Description
[Describe what this change achieves]
 
### Issues Resolved

### Check List


derek-ho:
@cwperks can you help take a look for technical accuracy? Also @pajuric I am not too familiar with the downloads page. Will removing this file remove the entry from the page?

nateynateynate:
Hey @derek-ho -  `_include/downloads/logstash-oss-with-opensearch-output-plugin.markdown` is referenced in many other locations. I don't think we can safely delete it unless we find the references to it.  

I forget where the logic is that constructs the view of the download page, but give me a bit and I'll drop some hints in here.

nateynateynate (changes requested):
It's always a good idea to try to build the site locally so we can see what build errors are going to pop up. The output of that will lead us to what's going wrong.

derek-ho:
@cwperks @dblock I am working on this PR but a little confused on how to proceed here. Should we remove the logstash in general, or only going forward?

pajuric:
@derek-ho - Do you still require feedback here? What can we do to assist you with closing this out?

PR #4230, opened 2026-09-18
Fix the 404 documentation links in the MCP blog post

### LINKED ISSUE #4001: [DOC] link for installing transport-reactor-netty4 plugin is a 404
**What do you want to do?**
 
- [x] Request a change to existing documentation
- [ ] Add new documentation
- [x] Report a technical problem with the documentation
- [ ] Other 

**Tell us about your request.** 

The `transport-reactor-netty4` documentation (listed as "step 1" in both) links fail in these two documents:
- https://opensearch.org/blog/introducing-mcp-in-opensearch/
- https://github.com/opensearch-project/project-website/blob/c896713b6e1e25add756d6e20583cb88fa05c558/_posts/2025-05-05-Introducing-MCP-in-OpenSearch.md#section-11-built-in-mcp-server

gohri81-tech:
### Description

Closes #4001.

The `transport-reactor-netty4` link in Step 1 of the MCP post pointed at

```
https://docs.opensearch.org/docs/latest/install-and-configure/configuring-opensearch/network-settings/#selecting-the-transport
```

The target page and the anchor both still exist - `_install-and-configure/configuring-opensearch/network-settings.md` in `documentation-website` has `## Selecting the transport` at line 293, and the docs site links to that exact anchor itself from `_api-reference/document-apis/bulk-streaming.md`. What is wrong is the path: on `docs.opensearch.org` the version comes straight after the host, so the `/docs` prefix left over from the old `opensearch.org/docs/latest/...` URLs makes the whole URL a 404.

So the fix is just to drop the duplicated segment:

```diff
- https://docs.opensearch.org/docs/latest/install-and-configure/configuring-opensearch/network-settings/#selecting-the-transport
+ https://docs.opensearch.org/latest/install-and-configure/configuring-opensearch/network-settings/#selecting-the-transport
```

Six other links in the same post have the identical problem, so all seven are fixed here rather than leaving the post half broken:

- `ml-commons-plugin/api/mcp-server-apis/index/` (MCP server APIs documentation)
- `install-and-configure/configuring-opensearch/network-settings/#selecting-the-transport` (the one in the issue)
- `ml-commons-plugin/agents-tools/tools/index/` (Tools)
- `ml-commons-plugin/agents-tools/agents/conversational/` (conversational agent)
- `ml-commons-plugin/agents-tools/agents/plan-execute-reflect/` (twice)
- `ml-commons-plugin/agents-tools/mcp/mcp-connector/` (full documentation)

Seven additions and seven deletions, nothing but the host path touched - the page paths and the `#selecting-the-transport` anchor are unchanged.

That the shorter form is the right one is visible in the repository itself: the posts written after the host moved use `https://docs.opensearch.org/latest/...`, for example `_posts/2025-11-19-introducing-agentic-search.md` and `_posts/2025-09-30-Introducing-Derived-Source.markdown`.

### Not in this PR

The same mistake is in 19 other posts, all from roughly April to August 2025, which is when the host changed. `repo:opensearch-project/project-website "docs.opensearch.org/docs/latest"` in code search lists 20 files in total, this one included. I have kept this PR to the post named in the issue so it stays reviewable, but I am happy to do the remaining 19 in a follow-up, or to add them here if you would rather land it in one go - just say which you prefer. A link checker rule on `_posts` would stop it coming back.

### Issues Resolved

Closes #4001

### Check List

- [x] Commits are signed per the DCO using `--signoff`
- Content only, no code or layout change

Two caveats, both about how I made the change rather than about the change itself. I cannot reach the docs site from where I am working, so I went by the URL shapes used elsewhere in this repository and by the file and heading in `documentation-website` rather than by clicking the links - please sanity check one of them before merging. And the web editor rewrote the final line of the file because it has no trailing newline, which is the last hunk in the diff with no visible change; happy to redo this locally without it if you would rather keep the diff to the seven links.


gohri81-tech:
A gentle nudge on this one when someone has a spare minute. It only repoints the documentation links in the MCP blog post that currently return 404, so the change is quick to verify by clicking the before and after URLs.

Happy to adjust the targets if you would rather they point somewhere else. Thank you.

